PR #68, opened 2023-04-16
Minor gas optimizations on reverts

yudilevi:


barakman (approved):
Nice!
The first change is actually an optimization for non-reverting transactions.

PR #90, opened 2023-05-30
When both rates are 0, prevent only trading

barakman:
The check `StrategyDoesNotExist` is based on the condition `if strategy storage slots are all 0`.

This implementation is incorrect, because the product definition allows creating such strategy, or updating an existing strategy to reach that state (for example, by withdrawing all the liquidity, and setting both rates on both orders to zero).

Implementing it in other ways would require more storage resources (thus more storage reads and writes).

On top of that, this check is conducted in every strategy-access context, i.e., trading on a strategy, updating a strategy and reading a strategy, while it is really only required in the context of trading.

Under this context, it can be done in a cheaper manner, by checking if both `A` and `B` of the target order are zero, inside each one of the trade calculation (pure) functions. Note that in contrast with the current check, this new check is per order and not for the entire strategy, and that the revert reason should subsequently change to `OrderDisabled`.

### DIFF
--- a/contracts/carbon/Strategies.sol
+++ b/contracts/carbon/Strategies.sol
@@ -137,7 +137,7 @@ abstract contract Strategies is Initializable {
     error InvalidRate();
     error InvalidTradeActionStrategyId();
     error InvalidTradeActionAmount();
-    error StrategyDoesNotExist();
+    error OrderDisabled();
     error OutDated();
 
     struct SourceAndTargetAmounts {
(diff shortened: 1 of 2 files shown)


yudilevi (approved):

barakman:
@yudilevi @yarivbancor @lbeder @ivanzhelyazkov :
Can you please review all execution paths of function `_unpackOrders`, while taking into account the fact that this function no longer asserts the "all slots zero" condition?

ivanzhelyazkov (approved):

PR #91, opened 2023-05-31
Validate deposit amounts

ivanzhelyazkov:


yudilevi (approved):

contracts/carbon/Strategies.sol
```diff
@@ -130,6 +130,7 @@ abstract contract Strategies is Initializable {
     using SafeCastUpgradeable for uint256;
 
     error NativeAmountMismatch();
+    error TokenAmountMismatch();
```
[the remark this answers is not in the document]
ivanzhelyazkov: renamed

contracts/carbon/Strategies.sol
```diff
@@ -635,7 +638,16 @@ abstract contract Strategies is Initializable {
                 payable(address(owner)).sendValue(txValue - depositAmount);
             }
         } else if (depositAmount > 0) {
-            token.safeTransferFrom(owner, address(this), depositAmount);
+            if (validateDepositAmount) {
+                uint256 balanceBefore = token.balanceOf(address(this));
```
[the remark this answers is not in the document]
ivanzhelyazkov: renamed

contracts/carbon/Strategies.sol
```diff
@@ -333,7 +334,7 @@ abstract contract Strategies is Initializable {
             } else if (newOrders[i].y > orders[i].y) {
                 // liquidity increased - deposit the difference
                 uint128 delta = newOrders[i].y - orders[i].y;
-                _validateDepositAndRefundExcessNativeToken(token, owner, delta, value);
+                _validateDepositAndRefundExcessNativeToken(token, owner, delta, value, true);
```
barakman, line 337: Technically speaking, we don't need it at this point, because the strategy creation would have failed.
It's not a 100% thing, since the token contract may have been upgraded to include transfer-fee (or perhaps even that fee mechanism "decides to not take a fee" on strategy creation).
But I'd say it's a 99.9999% thing, so your call...

contracts/carbon/Strategies.sol
```diff
@@ -635,7 +638,16 @@ abstract contract Strategies is Initializable {
                 payable(address(owner)).sendValue(txValue - depositAmount);
             }
         } else if (depositAmount > 0) {
-            token.safeTransferFrom(owner, address(this), depositAmount);
+            if (validateDepositAmount) {
+                uint256 prevBalance = token.balanceOf(address(this));
+                token.safeTransferFrom(owner, address(this), depositAmount);
+                uint256 newBalance = token.balanceOf(address(this));
+                if (newBalance - prevBalance != depositAmount) {
+                    revert BalanceMismatch();
```
barakman, line 646: Minor: the error-message, IMO, should be 'TransferFeeNotAllowed' or something like that.

contracts/carbon/Strategies.sol
```diff
@@ -481,7 +482,8 @@ abstract contract Strategies is Initializable {
             params.tokens.source,
             params.trader,
             totals.sourceAmount,
-            params.txValue
+            params.txValue,
+            false
```
barakman, line 486: Technically speaking, some sophisticated fee mechanism could detect that the function is called from create or update strategy, and avoid a fee only in those cases (which means that the fee will go undetected when trading).

In the previous comment, I've implied that we can do this check only when creating a strategy, in order to reduce strategy update gas cost.

In this comment, I'm implying that we should do it when creating a strategy, when updating a strategy and when trading, in order to ensure 100% no fee.

...

barakman (changes requested):

contracts/carbon/Strategies.sol
ivanzhelyazkov: Please discuss with Yudi to reach an agreement on that - his preference is BalanceMismatch, for me it also makes more sense.

contracts/carbon/Strategies.sol
ivanzhelyazkov: It's definitely possible to have fee-on-transfer activated later. Additionally, it's possible to have strategies with fee-on-transfer tokens created before the upgrade. imo it's worth it

contracts/carbon/Strategies.sol
ivanzhelyazkov: I've also mentioned this - for me it makes sense to have it as well, since some tokens may activate fees after a strategy has been created. The main issue here is the increased gas cost of trading - getting the balance of one token twice will cost 2100 (cold read) + 100 (hot read) = 2200 more gas.

barakman: Plus the external contract function call, which is another 3000 as far as I remember.
And you call it twice (before and after).
So that's 10k for one token, and there are possibly two tokens.

contracts/carbon/Strategies.sol
yudilevi: there's no to know the cause is a transfer fee

contracts/carbon/Strategies.sol
yudilevi: only a single token on trade but I'm reluctant to increase gas costs on trades for that

ivanzhelyazkov: We deposit one side, since it's a trade - so only up to one token per trade. Additionally, two external contract function calls cost about 2700 total: 2600 (cold access, base) for the first access + 100 for the second (warm access).
So total cost should be 2200 for sload + ~2700 for external call = ~5000

test/carbon/trading/trading.ts
```diff
@@ -303,6 +305,18 @@ describe('Trading', () => {
         await transfer(deployer, tokens[sourceSymbol], trader, expectedSourceAmount);
     };
 
+    /**
+     * generates a test order
+     */
+    const generateTestOrder = (): TestOrder => {
```
yarivbancor, line 311: this would be the 3rd copy of the same function, move to a shared utility function?

yarivbancor (changes requested):

test/carbon/trading/trading.ts
ivanzhelyazkov: exported into a shared utility function

barakman (approved):

contracts/helpers/TestERC20FeeOnTransfer.sol
```diff
@@ -0,0 +1,58 @@
...
+        _positiveFee = positiveFee;
+    }
+
+    /**
+     * @dev overriden transferFrom function which takes tax from every transfer
+     */
+    function transferFrom(address from, address to, uint256 amount) public override returns (bool) {
+        if (!_feeEnabled) {
+            return super.transferFrom(from, to, amount);
+        } else {
```
barakman, line 52: Minor: can get rid of the 'else'

barakman (approved):

barakman (approved):

contracts/helpers/TestERC20FeeOnTransfer.sol
ivanzhelyazkov: I think the path of execution can be seen clearer this way. Anyway, it's just a mock.

lbeder (approved):

test/helpers/Utils.ts
```diff
@@ -29,6 +29,17 @@ export const getBalance = async (token: TokenWithAddress, account: string | Addr
...
+export const getEvent = async (tx: ContractTransaction, eventName: string) => {
+    const events = (await tx.wait()).events;
+
+    const filteredEvents = events?.filter((e) => e.event === eventName);
+    if (filteredEvents === undefined) {
+        return [];
+    } else {
+        return filteredEvents;
+    }
+};
```
barakman, line 41: Minor:
```
    const events = (await tx.wait()).events || [];
    return events.filter((e) => e.event === eventName);
```

ivanzhelyazkov: addressed - better in this case is the ?? operator.

ivanzhelyazkov: addressed - better in this case is the ?? operator.

PR #95, opened 2023-06-02
Reduce trading gas cost

barakman:
Notes:
1. It is not possible to quantify the impact of each change independently, because these changes are tightly-coupled in terms of gas.
2. A few of these changes are actually in favor of reducing stack size, in order to allow for the other (gas-related) changes to be compilable.

### DIFF
--- a/contracts/carbon/Strategies.sol
+++ b/contracts/carbon/Strategies.sol
@@ -153,9 +153,13 @@ abstract contract Strategies is Initializable {
         uint128 constraint;
         uint256 txValue;
         Pair pair;
+        uint128 sourceAmount;
+        uint128 targetAmount;
     }
 
     uint256 private constant ONE = 1 << 48;
+
+    uint256 private constant ORDERS_INVERTED_FLAG = 1 << 255;
 
     uint32 private constant DEFAULT_TRADING_FEE_PPM = 2000; // 0.2%
 
@@ -307,7 +311,7 @@ abstract contract Strategies is Initializable {
     ) internal {
         // prepare storage variable
         uint256[3] storage packedOrders = _packedOrdersByStrategyId[strategyId];
-        uint256[3] memory packedOrdersMemory = _packedOrdersByStrategyId[strategyId];
+        uint256[3] memory packedOrdersMemory = packedOrders;
         (Order[2] memory orders, bool ordersInverted) = _unpackOrders(packedOrdersMemory);
 
         // revert if the strategy mutated since this tx was sent
@@ -317,10 +321,14 @@ abstract contract Strategies is Initializable {
 
         // store new values if necessary
         uint256[3] memory newPackedOrders = _packOrders(newOrders, ordersInverted);
-        for (uint256 n = 0; n < 3; n = uncheckedInc(n)) {
-            if (packedOrdersMemory[n] != newPackedOrders[n]) {
-                packedOrders[n] = newPackedOrders[n];
-            }
+        if (packedOrdersMemory[0] != newPackedOrders[0]) {
+            packedOrders[0] = newPackedOrders[0];
+        }
+        if (packedOrdersMemory[1] != newPackedOrders[1]) {
+            packedOrders[1] = newPackedOrders[1];
+        }
+        if (packedOrdersMemory[2] != newPackedOrders[2]) {
+            packedOrders[2] = newPackedOrders[2];
         }
 
         // deposit and withdraw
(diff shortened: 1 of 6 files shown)


contracts/carbon/Strategies.sol
```diff
@@ -390,118 +398,130 @@ abstract contract Strategies is Initializable {
...
-                order1: orders[1],
-                reason: STRATEGY_UPDATE_REASON_TRADE
-            });
-
-            totals.sourceAmount += tempTradeAmounts.sourceAmount;
-            totals.targetAmount += tempTradeAmounts.targetAmount;
+            if (ordersInverted) {
+                emit StrategyUpdated({
+                    id: strategyId,
+                    token0: params.pair.tokens[1],
```
lbeder, line 462: You should consider using `ordersInverted` inline (i.e., `token0: params.pair.tokens[ordersInverted ? 1 : 0]`) to reduce the impact on the size of the contract.

lbeder (commented):
See a preliminary comment below. I'd still need to keep reviewing the PR.

lbeder (commented):
See a preliminary comment below. I'd still need to keep reviewing the PR.

contracts/carbon/Strategies.sol
barakman: That would yield 2 branches (for token0 and for token1) **per order** (since it's inside the 'for' loop).
If I remember correctly, it doesn't impact the contract size significantly.

lbeder: AFAIK, the "duplication" of the whole `emit` is worse, but we should measure and compare in any case.

barakman: I'll do that

barakman: 23.133 with two "emits".
23.100 with one "emit".

lbeder:
Ah, nm then…

contracts/carbon/Strategies.sol
yudilevi: same gas cost, so I'd keep a single branch for readability

contracts/carbon/Strategies.sol
```diff
@@ -317,10 +321,14 @@ abstract contract Strategies is Initializable {
 
         // store new values if necessary
         uint256[3] memory newPackedOrders = _packOrders(newOrders, ordersInverted);
-        for (uint256 n = 0; n < 3; n = uncheckedInc(n)) {
-            if (packedOrdersMemory[n] != newPackedOrders[n]) {
-                packedOrders[n] = newPackedOrders[n];
-            }
+        if (packedOrdersMemory[0] != newPackedOrders[0]) {
```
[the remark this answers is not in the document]
barakman: Removing this specific optimization increases the cost by 532 gas per order (1064 for 2 orders, 5320 for 10 orders, etc). Relatively to the other changes, this one is among the more significant ones.

contracts/carbon/Strategies.sol
barakman: This specific change even increases the gas cost (very slightly).
It is actually in favor of reducing stack size, in order to allow for the other (gas-related) changes to be compilable.
Otherwise, you get stack-too-deep, which I couldn't find any way to resolve other than with `viaIR: true`.
Can probably resolve by moving into a separate function, if you wish.

contracts/carbon/Strategies.sol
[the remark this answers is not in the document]
yudilevi: Ok, can live with it

contracts/carbon/Strategies.sol
barakman: Cannot resolve with a separate function.

yudilevi: Any other alternatives?

lbeder: What are you trying to do? You don't have to use new variables for that, but can use the ternary `? :` instead, no?

barakman: Option 1:
Get rid of most of the other optimizations in this function, specifically where more local variables are used, for example, getting rid of `uint128 amount = tradeActions[i].amount` AND reverting `_singleTradeActionSourceAndTargetAmounts` to return a memory struct instead of two variables. This would revert a  quite-significant gas reduction.

Option 2:
Check the condition twice instead of once:
```
emit StrategyUpdated({
    id: strategyId,
    token0: params.pair.tokens[ordersInverted ? 1 : 0],
    token1: params.pair.tokens[ordersInverted ? 0 : 1],
    order0: orders[0],
    order1: orders[1],
    reason: STRATEGY_UPDATE_REASON_TRADE
});
```
Which increases the gas cost by either 92 or 140 (depending on the type of trade) gas per order.

Option 3:
Change the event prototype to take one `Token[2]` parameter instead of two `Token` parameters.
This is an API-breaking change of course.

Option 4:
Reduce the event even further, and remove these two tokens altogether, since they can be retrieved also from the strategy ID itself (which contains the pool ID).
This is also an API-breaking change of course.

barakman: > What are you trying to do? You don't have to use new variables for that, but can use the ternary `? :` instead, no?

Correct. This would yield two checks of `ordersInverted` instead of one (which is the current implementation on mainnet). See the impact of this change in the comment above.

barakman: Proceeded with option 2.
Please review.

yudilevi (approved):

yarivbancor (approved):

yudilevi (approved):

yudilevi (approved):

ivanzhelyazkov (approved):

PR #98, opened 2023-06-08
Improve code coverage

ivanzhelyazkov:


test/carbon/Pairs.ts
```diff
@@ -1,19 +1,27 @@
...
 
 describe('Pairs', () => {
     let carbonController: CarbonController;
+    let emergencyStopper: SignerWithAddress;
     let token0: TestERC20Burnable;
     let token1: TestERC20Burnable;
     let pair: any;
 
+    before(async () => {
+        [, , emergencyStopper] = await ethers.getSigners();
```
barakman, line 22: Minor: can probably get rid of one of the commas here

test/carbon/Strategies.ts
```diff
@@ -1427,6 +1427,23 @@ describe('Strategy', () => {
                 );
         });
 
+        it('should be able to delete a disabled order', async () => {
```
barakman, line 1430: A more accurate description is `should be able to delete a strategy when both orders are disabled`.
I would also explicitly create such strategy (I think that you are creating it implicitly here).

Perhaps even split this test into 3 tests:
- 1st order disabled, 2nd order enabled
- 1st order enabled, 2nd order disabled
- 1st order disabled, 2nd order disabled

barakman (approved):
A couple of minor comments, so approving...

lbeder (approved):
IMHO, it would be nice to add the coverage to the README.md (similar to V3's https://github.com/bancorprotocol/contracts-v3/tree/dev#test-coverage)

lbeder (approved):
IMHO, it would be nice to add the coverage to the README.md (similar to V3's https://github.com/bancorprotocol/contracts-v3/tree/dev#test-coverage)

test/carbon/Pairs.ts
ivanzhelyazkov: well we need the third signer only, and we don't need the first two, so if I remove a comma, it's going be the incorrect signer.

test/carbon/Strategies.ts
ivanzhelyazkov: added

ivanzhelyazkov:
> IMHO, it would be nice to add the coverage to the README.md (similar to V3's https://github.com/bancorprotocol/contracts-v3/tree/dev#test-coverage)

added to readme

test/carbon/Strategies.ts
```diff
@@ -86,7 +86,7 @@ describe('Strategy', () => {
         // prepare variables
         const _params = { ...params };
         const order = _params.order ? _params.order : generateTestOrder();
-        const secondOrder = order ? order : generateTestOrder();
+        const secondOrder = order || generateTestOrder();
```
lbeder, line 89: Nitpick: I usually prefer `??` (over `||`) because it won't override falsy values (e.g., `0` or `""`), but only undefined/optional values

lbeder (approved):

test/carbon/Strategies.ts
ivanzhelyazkov: fixed

yarivbancor (approved):

PR #105, opened 2023-06-27
Carbon vortex

ivanzhelyazkov:
* Add changes for Fee Burner (now called Carbon Vortex) contract:
* Token balances are accounted for as well as Carbon fees
* Amount swapped to BNT for each token is only up to `100 - rewards %`
* User receives rewards in each of the tokens sent to the `execute` function, instead of only in BNT


contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+    using Address for address payable;
+
+    ICarbonController private immutable _carbonController;
+    IBancorNetwork private immutable _bancorNetwork;
+    Token private immutable _bnt;
+
+    uint256 private _totalBurnt;
+
+    // rewards percentage
+    uint256 private _rewardsPercentagePPM;
```
barakman, line 42: `uint32 private _rewardsPPM;`

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+     * @inheritdoc Upgradeable
+     */
+    function version() public pure override(IVersioned, Upgradeable) returns (uint16) {
+        return 2;
+    }
+
+    /**
+     * @inheritdoc ICarbonVortex
+     */
+    function setRewards(uint256 newRewardsPercentagePPM) public onlyAdmin validFee(uint32(newRewardsPercentagePPM)) {
```
barakman, line 101: `function setRewardsPPM(uint32 newRewardsPPM) public onlyAdmin validFee(newRewardsPPM)`

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+        uint256[] memory rewardAmounts = new uint256[](len);
+
+        // withdraw fees, load balances and reward amounts
+        for (uint256 i = 0; i < len; i = uncheckedInc(i)) {
+            // withdraw token fees
+            _carbonController.withdrawFees(tokens[i], type(uint256).max, address(this));
+            // get token balance
+            balances[i] = tokens[i].balanceOf(address(this));
+            // get reward amount for token
+            rewardAmounts[i] = MathEx.mulDivF(balances[i], _rewardsPercentagePPM, PPM_RESOLUTION);
```
barakman, line 153: Cache `_rewardsPercentagePPM` in order to avoid reading a storage slot on every iteration

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        // convert tokens to BNT
+        for (uint256 i = 0; i < len; i = uncheckedInc(i)) {
```
barakman, line 157: Why do we need to split that into two loops?
Moreover, why do we need to build the `balances` array when only a single balance is relevant at each iteration, and the array itself isn't used outside the loop?

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+            Token token = tokens[i];
+            // skip token if no token balance
+            if (balances[i] == 0) {
+                continue;
+            }
+            // check if token is BNT - we don't need to swap in this case
+            if (token == _bnt) {
```
barakman, line 164: Can use `if (balances[i] == 0 || token == _bnt)`

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+                continue;
+            }
+            if (token.isNative()) {
+                // using a regular transfer here could revert due to exceeding the 2300 gas limit
+                // which is why we're using call instead (via sendValue)
+                payable(sender).sendValue(rewardAmount);
+            } else {
+                token.safeTransfer(sender, rewardAmount);
+            }
+        }
```
barakman, line 203: Maybe do this as part of the loop in function `execute`

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+        // get the burn amount
+        uint256 burnAmount = _bnt.balanceOf(address(this));
+
+        // add to the total burnt amount
+        if (burnAmount > 0) {
+            _totalBurnt += burnAmount;
+        }
+
+        // burn the tokens
+        _bnt.safeTransfer(Token.unwrap(_bnt), burnAmount);
```
barakman, line 214: This can be inside the `if (burnAmount > 0)` I believe

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+     */
+    modifier validateTokens(Token[] calldata tokens) {
+        uint len = tokens.length;
+        if (len == 0) {
+            revert InvalidTokenLength();
+        }
+        for (uint256 i = 0; i < len; i = uncheckedInc(i)) {
+            Token token = tokens[i];
+            // validate token has no duplicates
+            for (uint256 j = i + 1; j < len; j = uncheckedInc(j)) {
```
barakman, line 250: `uint256 j = uncheckedInc(i)`

barakman (changes requested):

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+
+        _rewardsPercentagePPM = newRewardsPercentagePPM;
+
+        emit RewardsUpdated({ prevRewards: prevRewardsPercentagePPM, newRewards: newRewardsPercentagePPM });
+    }
+
+    /**
+     * @inheritdoc ICarbonVortex
+     */
+    function rewards() external view returns (uint256) {
```
barakman, line 117: `function rewardsPPM() external view returns (uint32)`

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+        return _rewardsPercentagePPM;
+    }
+
+    /**
+     * @inheritdoc ICarbonVortex
+     */
+    function totalBurnt() external view returns (uint256) {
```
barakman, line 124: Cosmetic: I believe that 'burned' will be consistent with the rest of the code

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+                // using a regular transfer here could revert due to exceeding the 2300 gas limit
+                // which is why we're using call instead (via sendValue)
+                payable(sender).sendValue(rewardAmount);
+            } else {
+                token.safeTransfer(sender, rewardAmount);
+            }
+        }
+
+        // get the burn amount
+        uint256 burnAmount = _bnt.balanceOf(address(this));
```
barakman, line 206: I believe that you can avoid this external call, and instead, accumulate the actual burn amount by summing up the values returned from the call to `_bancorNetwork.tradeBySourceAmount`.
In fact, it might even be the more correct thing to do, since the contract may be holding BNT for some different reason altogether (i.e., before this transaction is executed).

barakman (changes requested):

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Consider if it was one loop, in which we get the balance, user rewards and swap
the case when a user sends [TKN1, TKN2, BNT] : 
1. we get the token balance of TKN1 and allocate 2% user rewards for TKN1, swap 98% of TKN1 for BNT
2. we get the token balance of TKN2 and allocate 2% user rewards for TKN2, swap 98% of TKN2 for BNT
3. we get the token balance of BNT - which has increased based on the last two swaps significantly, and allocate 2% user rewards in BNT
You can see how the user can inflate his rewards by reordering the tokens he sends.
By using two loops, we make sure the user doesn't receive more rewards than he should.
We cache the balances so that we don't call `balanceOf` again in the second loop.

barakman: Are you saying that the call to `_bancorNetwork.tradeBySourceAmount`  impacts the outcome of the next call to `tokens[i].balanceOf(address(this))`?

ivanzhelyazkov: Of course, since we trade the current token for BNT - so, the BNT balance of the contract increases.
Also, the current token balance decreases.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Only case I see where this makes sense if the contract has BNT balance before, but the user doesn't send BNT as a token input, only other tokens - so the user in this case won't receive his 2% of the BNT burnt. Have to think a bit more about this case, it might make sense to accumulate the burn amount.

contracts/vortex/CarbonVortex.sol
barakman: But you are only using the balance of that token in the context of one specific iteration.
So how could `tradeBySourceAmount(bnt, tokens[i])` impact `tokens[i + 1].balanceOf(address(this))`?

contracts/vortex/CarbonVortex.sol
barakman: Well, you save both the function call (to `safeTransfer`), and the additional check (of `burnAmount > 0`) inside that function.

contracts/vortex/CarbonVortex.sol
barakman: Yup, that's the only case that I had in mind as well. And there's a good question of whether or we should include that additional balance as part of the process.
But regardless - my main point was to use information that you already have (the return-value of `tradeBySourceAmount`) in order to avoid the external function call (to `balanceOf` on the BNT contract).

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,262 @@
...
+        __ReentrancyGuard_init();
+
+        __CarbonVortex_init_unchained();
+    }
+
+    /**
+     * @dev performs contract-specific initialization
+     */
+    function __CarbonVortex_init_unchained() internal onlyInitializing {
+        setRewards(100_000);
```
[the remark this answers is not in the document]
ivanzhelyazkov: I don't think that's really important - the contract has been initialized already.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Hmm, I think it's better to leave it as it is - it becomes difficult to track the logic flow otherwise imo

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: changed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: changed (keep in mind it should be uint256 so as not to corrupt the next possible slot during upgrade)

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: changed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: updated

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: cached

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: updated

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: updated (keep in mind it should be uint256 so as not to corrupt the next storage slot during an upgrade)

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: It does impact the next token if that token is BNT

barakman: It is never BNT as far as I can see, because of the `if (token == _bnt) continue;` line a little bit further down the code.
But perhaps I'm missing some context here, so this is for you to consider (a single for loop, without the need to maintain an array of all the balances).

contracts/vortex/CarbonVortex.sol
barakman: It's a new contract (first-time deployment, not an upgrade); how can any storage be corrupted here?

contracts/vortex/CarbonVortex.sol
barakman: There are still a ton of 'burnt' in the code (events, documentation, etc).
Perhaps a find/replace would be better...

contracts/vortex/CarbonVortex.sol
barakman: It's a new contract (first-time deployment, not an upgrade); how can any storage be corrupted here?

contracts/vortex/CarbonVortex.sol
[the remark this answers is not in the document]
barakman: More of a matter of convention I suppose (in an attempt to have as few public functions as possible).
There's also the redundant check of `adminOnly`, which takes place here.
Both of these aspects are minor of course, so leaving this comment here for a short while...

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: it will be upgraded from `FeeBurner`, the name is changed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: The user rewards for BNT will be calculated before that check - since even if we don't swap, user still needs to receive his BNT rewards - where the entire issue comes from.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: it will be upgraded from FeeBurner, the name is changed

contracts/vortex/CarbonVortex.sol
barakman: So this:
```
uint256 private _totalBurnt;
uint32 percentagePPM;
uint224 maxAmount;
```
Changes to this:
```
uint256 private _totalBurned;
uint256 private _rewardsPPM;
```
Which makes it fine to declare `_rewardsPPM` as `uint32`.

contracts/vortex/CarbonVortex.sol
barakman: See [my previous comment](https://github.com/bancorprotocol/carbon-contracts/pull/105#discussion_r1244832819)

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: I get what you mean, but there are a lot of cases to consider here and there is a balance of keeping the code simpler and introducing new variables.
Consider that BNT gets passed as a token - we don't swap it - how are we going to get the BNT burn amount if we don't check the BNT balance ?
The return value of `tradeBySourceAmount` is not enough in this case, so we'll need to make another balanceOf call, and then sum that with the total accumulated burn amount. 
So if we don't want to make two calls to `balanceOf`, we're going to have to pass two variables to `_allocateRewards` (burn accumulator and bnt balance, which may be 0) just for this case.
I believe the current approach is simpler and any user who uses the vortex will want to maximize his rewards, thus if there is token balance of any token, he will send that token as well to the call.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Saves about 50 gas, but it's still something. Changed.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Changed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: It's not fine - if we introduce another variable, it may be packed in the same slot (if it's <= uint224), which will be corrupted with data from maxAmount.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: It's not fine - if we introduce another variable, it may be packed in the same slot (if it's <= uint224), which will be corrupted with data from maxAmount.
I don't see an issue declaring rewardsPPM as uint256 as well.

contracts/vortex/CarbonVortex.sol
barakman: The return value of `tradeBySourceAmount` is of course not relevant for BNT, because we don't call that function in this case. We can just add the BNT amount to the total accumulator in this case. My point is to get rid of the `_bnt.balanceOf(this)` call at the end of the function, both for gas, as well as in order to really consider only the amount which was traded (excluding a potential previous balance).
Anyway, closing...

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,259 @@
...
+    using Address for address payable;
+
+    ICarbonController private immutable _carbonController;
+    IBancorNetwork private immutable _bancorNetwork;
+    Token private immutable _bnt;
+
+    uint256 private _totalBurned;
+
+    // rewards percentage
+    uint256 private _rewardsPPM;
```
lbeder, line 42: I do think it's a bit dangerous leaving it like this, since we might forget in the future that this is the result of porting `Rewards` to `_rewardsPPM`. What about keeping it as `uint32`, while declaring a `uint224 _reserved0` instead which can be set to `0` during `_postUpgrade` and deleted in a future upgrade?

lbeder: Alternatively, we can just deploy a new `CarbonVortex` instance (i.e., with `version` 1) and ignore the prior instance.

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,259 @@
...
+    function uncheckedInc(uint256 i) private pure returns (uint256 j) {
+        unchecked {
+            j = i + 1;
+        }
+    }
+
+    /**
+     * @dev perform various validations for the token array
+     */
+    modifier validateTokens(Token[] calldata tokens) {
```
lbeder, line 239: Nitpick> In the past, we have been usually defining modifiers at the top and moving its logic to a helper function.

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,259 @@
...
+        uint256[] memory balances = new uint256[](len);
+        // allocate array for the reward amounts for caller
+        uint256[] memory rewardAmounts = new uint256[](len);
+        // cache rewardsPPM to save gas
+        uint256 rewardsPercentage = _rewardsPPM;
+
+        // withdraw fees, load balances and reward amounts
+        for (uint256 i = 0; i < len; i = uncheckedInc(i)) {
+            // withdraw token fees
+            _carbonController.withdrawFees(tokens[i], type(uint256).max, address(this));
```
lbeder, line 151: Future (?) gas optimization: `CarbonController` can expose a `withdrawFees` variant that allows withdrawing multiple token fees in the same call (thus saving all the external calls in the loop).

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,259 @@
...
+            // get token balance
+            balances[i] = tokens[i].balanceOf(address(this));
```
lbeder, line 153: `withdrawFees` returns the withdrawn amount, so perhaps it can be used to avoid calling `balanceOf`. It won't handle the case of "accidentally stuck" tokens, but:
1. It'll decrease gas costs nicely.
2. These can be handled by a dust collecting function that can be added to the vortex in the future.

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,259 @@
...
+        // convert tokens to BNT
+        for (uint256 i = 0; i < len; i = uncheckedInc(i)) {
+            Token token = tokens[i];
+            // skip token if no token balance or token is BNT - no need to swap in this case
+            if (balances[i] == 0 || token == _bnt) {
+                continue;
+            }
+
+            // get trade amount for token
+            uint256 tradeAmount = balances[i] - rewardAmounts[i];
```
lbeder, line 167: This operation can be `unchecked`

contracts/vortex/CarbonVortex.sol
```diff
@@ -0,0 +1,259 @@
...
+
+            // approve tokens for trading on Bancor Network V3
+            _setAllowance(token, tradeAmount);
```
lbeder, line 170: Instead of increasing the allowance, which is currently resulting in two `allowance` external calls (one here and one in `SafeERC20`'s `safeIncreaseAllowance`) and an `approve` external call you should consider (mutually exclusive):

1. Remove the first redundant `allowance` call
2. Set the allowance to `type(uint256).max` only once by either:
a. Have a global (not even permissioned) `approveBancorNetwork` function which will approve `_bancorNetwork` for a token/list of tokens.
b. Have a mapping of which tokens were already approved and check and set it once (it's less gas efficient than the previous option, but removes the need for a global function to handle this).

BTW, please note that the approach won't work for tokens (such as USDT) which require setting the approval first to 0, in case that for some reason the approval wasn't exact.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Actually the reason we use balanceOf is that the vortex will be used to store the rewards coming from the fast lane. They will be burned through the vortex.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Why is it dangerous ? 
We'll change the _rewardsPPM value during the upgrade, and this way it'll cover the entire slot. We'll have the proper `_rewardsPPM` value, so the rewards will be correct, and we won't have to think about this in future upgrades.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Actually, we do two calls only on the first approval for a token - after it will be only the `allowance` call.
`uint256` is a huge number and for most tokens, during the lifetime of the contract, that will be enough.
So the only extra call will be to `allowance` for each token. 
While I agree it will save more gas to have a centralized function to approve the tokens, there may be new tokens in the future not supported on V3, and there may be more changes, so this approach is most flexible at the overhead of one external call.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: updated

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Could you share why it worries you that we might forget it ?
The variable itself will be set in the `_postUpgrade` callback, meaning it would immediately become the proper value during upgrade.
Using `uint256` is cleaner in my opinion and since we'll set the value as the new `uint256`, we won't have issues with any dirty upper bits. Also it will cover the entire slot and won't corrupt the storage during a future upgrade.
If you mean to reserve the `uint224` slot in case we decide to use it in the future, it may make sense in this context.

contracts/vortex/CarbonVortex.sol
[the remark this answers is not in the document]
ivanzhelyazkov: Replaced the `setRewards` with an internal function call.

barakman (approved):

contracts/vortex/CarbonVortex.sol
lbeder: We discussed this offline. The TL;DR is that there are two options:
1. Keep it as is and add a `_postUpgrade` callback to this PR. 
2. Change it to `uint32`, so that the rest of the code will become similar to how `PPM`s are usually handled, while adding a reserved `uint224` (which will be also set in the `_postUpgrade` callback).

I'm totally fine with whatever @ivanzhelyazkov decides.

contracts/vortex/CarbonVortex.sol
lbeder: Sure thing

contracts/vortex/CarbonVortex.sol
lbeder: IMHO, managing allowances externally would be still better especially to the reasons that you've mentioned (would need to support other tokens, other approval targets, etc.), but it can be done later.

Another (more complicated) alternative is to actually try to do the trade without an approval and `try/catch` the revert from V3 and only then `approve` and try again. This will be more complicated to code, but will also result in no `allowance` calls in a happy flow.

deploy/scripts/0006-CarbonVortex-upgrade.ts
```diff
@@ -0,0 +1,20 @@
...
+import { DeployFunction } from 'hardhat-deploy/types';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+
+const func: DeployFunction = async ({ getNamedAccounts }: HardhatRuntimeEnvironment) => {
+    const { deployer, bnt, bancorNetworkV3 } = await getNamedAccounts();
+
+    const carbonController = await DeployedContracts.CarbonController.deployed();
+
+    await upgradeProxy({
+        name: InstanceName.CarbonVortex,
```
lbeder, line 11: This might not work if the "instance name" was perviously "FeeBurner", no?

ivanzhelyazkov: I've renamed all instances of FeeBurner with CarbonVortex (even in deployments) - it was the easier approach than keeping a legacy FeeBurner interface and contract.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: I think it could be worth it to do the external allowance management at a later point. 
In terms of gas, it will save 2100 gas per token in the call. Still, it will require tracking of the tokens which are going to be supported and manually approving them, which is a bit of operational overhead.

The second approach is rather interesting, I could consider implementing it.
It'd be good to hear some other opinions on this.

yudilevi: I’d keep it simple

deploy/scripts/0006-CarbonVortex-upgrade.ts
lbeder: Ah, gotcha!

lbeder (approved):

yudilevi (approved):

yudilevi (approved):

yudilevi (approved):

yudilevi (approved):

contracts/vortex/interfaces/ICarbonVortex.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: BUSL-1.1
-pragma solidity ^0.8.0;
+pragma solidity 0.8.19;
```
lbeder, line 2: In general, it's better to have a flexible `pragma` in interfaces (and common definitions), so that we won't force external integrations to use a specific solidity version.

lbeder (approved):

PR #107, opened 2023-07-18
Forge tests

ivanzhelyazkov:
* Add Forge configuration in Carbon repo
* Rewrite tests using Forge
* Configure CI to use Forge tests


test/forge/CarbonVortex.sol
```diff
@@ -0,0 +1,644 @@
...
+
+import { Token, toIERC20, NATIVE_TOKEN } from "../../contracts/token/Token.sol";
+
+contract CarbonVortexTest is TestFixture {
+    using Address for address payable;
+
+    address private bancorNetworkV3;
+
+    uint256 private constant REWARDS_PPM_DEFAULT = 100_000;
+    uint256 private constant REWARDS_PPM_UPDATED = 110_000;
```
barakman, line 28: Minor: we use `uint32` for all PPM values.

test/forge/Strategies.sol
```diff
@@ -0,0 +1,1829 @@
...
+        Token indexed token1,
+        Order order0,
+        Order order1,
+        uint8 reason
+    );
+
+    /**
+     * @dev Emitted when `tokenId` token is transferred from `from` to `to`.
+     */
+    event Transfer(address indexed from, address indexed to, uint256 indexed tokenId);
```
barakman, line 85: Can we import all of these events instead of redeclaring them?

test/forge/Strategies.sol
```diff
@@ -0,0 +1,1829 @@
...
+        uint256 val = tokens[i0] == NATIVE_TOKEN || tokens[i1] == NATIVE_TOKEN ? order.y : 0;
+        uint256 strategyId = carbonController.createStrategy{ value: val }(tokens[i0], tokens[i1], [order, order]);
+
+        // create new orders
+        Order memory order0 = updateOrderDelta(order, deltas[0]);
+        Order memory order1 = updateOrderDelta(order, deltas[1]);
+
+        // update strategy
+        val = tokens[i0] == NATIVE_TOKEN && deltas[0] >= 0 ? abs(deltas[0]) : 0;
+        val += tokens[i1] == NATIVE_TOKEN && deltas[1] >= 0 ? abs(deltas[1]) : 0;
```
barakman, line 503: You can get rid of the `abs` (you'll probably need to cast to uint though, but it would be safe to do so without using `safeCast`).

This code appears in many other places, so consider making it a custom helper function.

test/forge/Strategies.sol
```diff
@@ -0,0 +1,1829 @@
...
+        uint256 valueToChange
+    ) public {
+        vm.startPrank(user1);
+        valueToChange = bound(valueToChange, 0, 3);
+
+        Order memory order = generateTestOrder();
+        // create strategy
+        uint256 strategyId = carbonController.createStrategy(token0, token1, [order, order]);
+
+        if (valueToChange == 0) {
```
barakman, line 733: I suggest passing an array of delta values as input.
You could then pass, for example `[0, 1, 0, 0]`, which would mimic what you're doing here.
But it will allow you to extend this test in the future, and pass an array of delta values which isn't necessarily made of one "1" and three "0".
In fact, you can even use an `Order` struct for these delta values, making it clear which delta value is designated for which field.

barakman (changes requested):
Partially reviewed; I believe that you've pushed `testData2.json` by mistake.

test/forge/TestCaseParser.sol
```diff
@@ -0,0 +1,341 @@
...
+            } else {
+                (strategies, tradeActions, sourceAmount, targetAmount) = parseTestDataTemplate(
+                    "testCaseTemplateBySourceAmount"
+                );
+            }
+        }
+        if (inverseOrders) {
+            // swap orders for some strategies if true
+            for (uint256 i = 0; i < strategies.length; ++i) {
+                if (i % 2 == 0) {
```
barakman, line 163: You can just use `i += 2` in the `for` loop (then get rid of this check).

test/forge/TestCaseParser.sol
```diff
@@ -0,0 +1,341 @@
...
+        }
+        return result;
+    }
+
+    /// @dev helper function to convert an uint256 to string
+    function uintToString(uint256 v) private pure returns (string memory) {
+        if (v == 0) return "0";
+
+        uint256 maxlength = 100;
+        bytes memory reversed = new bytes(maxlength);
```
barakman, line 328:
```
bytes[100] memory reversed;
```

test/forge/TestCaseParser.sol
```diff
@@ -0,0 +1,341 @@
...
+            v = v / 10;
+            reversed[i++] = bytes1(uint8(48 + remainder));
+        }
+        bytes memory s = new bytes(i);
+        for (uint256 j = 0; j < i; j++) {
+            s[j] = reversed[i - 1 - j];
+        }
+        return string(s);
+    }
+}
```
barakman, line 341: Simplification suggestion:
```
function uintToString(uint256 x) private pure returns (string memory) {
    if (x > 0) {
        string memory str;
        while (x > 0) {
            str = string(abi.encodePacked(uint8(x % 10 + 48), str));
            x /= 10;
        }
        return str;
    }
    return "0";
}
```
Alternatively, consider importing these helper functions from [openzeppelin](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Strings.sol).

barakman (changes requested):

test/forge/CarbonVortex.sol
ivanzhelyazkov: Actually, not for CarbonVortex, due to a recent upgrade replacing the rewards storage struct with rewards ppm variable.

test/forge/Strategies.sol
ivanzhelyazkov: Only in solidity 0.8.21

test/forge/TestCaseParser.sol
ivanzhelyazkov: Have to allocate the array dynamically in order to be able to populate it like so:
`reversed[i++] = bytes1(uint8(48 + remainder));`

test/forge/Strategies.sol
ivanzhelyazkov: Extracted the logic in a separate function, casted directly to uint64 instead of using abs.

test/forge/Strategies.sol
ivanzhelyazkov: Replaced the `valueToChange` with an Order struct and simply added the deltas - also increased the number of values for the deltas from 0 to 100 000.

test/forge/TestCaseParser.sol
ivanzhelyazkov: Implemented

test/forge/TestCaseParser.sol
ivanzhelyazkov: Implemented

test/forge/TestCaseParser.sol
ivanzhelyazkov: The bytes1 casting doesn't work using the static allocation. Anyway, I implemented the approach you suggested below.

barakman (approved):

foundry.toml
```diff
@@ -0,0 +1,12 @@
+[profile.default]
```
lbeder, line 1: You should consider adding a profile with the same optimization settings as the deployed contracts so that it'd be possible to run the tests against the same version

package.json
```diff
@@ -24,10 +24,11 @@
   "scripts": {
     "build": "hardhat compile",
     "rebuild": "yarn clean && yarn build",
-    "test": "NODE_OPTIONS=${CI:+'--max-old-space-size=8192'} hardhat test",
+    "test": "forge test",
+    "test:ts": "NODE_OPTIONS='--max-old-space-size=8192' hardhat test",
```
lbeder, line 28: Perhaps `test:hardhat`?

test/forge/CarbonController.sol
```diff
@@ -0,0 +1,183 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.19;
```
lbeder, line 2: Perhaps `CarbonController.t.sol`?

package.json
```diff
@@ -24,10 +24,11 @@
...
     "test:coverage": "yarn build && NODE_OPTIONS='--max-old-space-size=8192' hardhat coverage",
     "test:coverage:nightly": "NIGHTLY=1 yarn test:coverage",
-    "test:ci": "CI=1 yarn test",
+    "test:ci": "CI=1 yarn test:ts",
```
lbeder, line 31: It `test:ci` even needed at this point?

foundry.toml
ivanzhelyazkov: Set the optimizer runs in the default profile to the same as the hardhat configuration.

package.json
ivanzhelyazkov: changed

test/forge/CarbonController.sol
ivanzhelyazkov: Replaced all forge test filenames to ones with the same format.

test/forge/CarbonController.t.sol
```diff
@@ -0,0 +1,183 @@
...
+        assertEq(version, 2);
+
+        bytes32 adminRole = keccak256("ROLE_ADMIN");
+        bytes32 emergencyStopperRole = keccak256("ROLE_EMERGENCY_STOPPER");
+        bytes32 feesManagerRole = keccak256("ROLE_FEES_MANAGER");
+        assertEq(adminRole, carbonController.roleAdmin());
+        assertEq(emergencyStopperRole, carbonController.roleEmergencyStopper());
+        assertEq(feesManagerRole, carbonController.roleFeesManager());
+
+        assertEq(admin, carbonController.getRoleMember(adminRole, 0));
```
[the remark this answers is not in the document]
ivanzhelyazkov: Added tests for role members count in CarbonController and Voucher.

test/forge/CarbonController.t.sol
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * @dev pausing / unpausing
+     */
+
+    /// @dev test emergency stopper should be able to pause
+    function testShouldBeAbleToPause() public {
+        vm.prank(emergencyStopper);
+        vm.expectEmit();
+        emit Paused(emergencyStopper);
```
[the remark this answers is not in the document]
ivanzhelyazkov: Added state tests as well.

test/forge/CarbonController.t.sol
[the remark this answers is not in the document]
yudilevi: what about the other roles?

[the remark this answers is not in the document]
ivanzhelyazkov: added

yudilevi (approved):

lbeder (approved):

yudilevi (approved):

barakman (approved):

lbeder (approved):

deploy/tests/carbon.ts
```diff
@@ -306,7 +313,7 @@ import { ethers, getNamedAccounts } from 'hardhat';
             for (let i = 0; i < strategies.length; ++i) {
                 const strategy = strategies[i];
                 // encode orders to values expected by the contract
-                const orders = strategy.orders.map((order) =>
+                const orders = strategy.orders.map((order: any) =>
```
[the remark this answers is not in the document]
barakman: Maybe something like:
```
order: {liquidity: String, lowestRate: String, highestRate: String, marginalRate: String}
```

PR #109, opened 2023-07-21
Custom taker fees

ivanzhelyazkov:
* Add custom taker fees (custom fees per pair) implementation
* Add custom taker fees tests


barakman (commented):
General naming suggestion:

I'd use `_defaultTradingFeePPM` and `_customTradingFeePPM` or something like that.

contracts/carbon/CarbonController.sol
```diff
@@ -122,6 +122,13 @@ contract CarbonController is
         return _tradingFeePPM;
     }
 
+    /**
+     * @inheritdoc ICarbonController
+     */
+    function tradingFeePPMOverrides(uint128 pairId) external view returns (uint32) {
+        return _tradingFeePPMOverrides[pairId];
```
lbeder, line 129: Is override the right name for this? It could be just something like "custom fees" or even just "fees" (and then a default fee for everything that isn't defined explicitly).

contracts/carbon/Strategies.sol
```diff
@@ -178,14 +178,22 @@ abstract contract Strategies is Initializable {
     // accumulated fees per token
     mapping(Token => uint256) internal _accumulatedFees;
 
+    // mapping between a pair id to its custom trading fee (in units of PPM)
+    mapping(uint128 => uint32) internal _tradingFeePPMOverrides;
```
lbeder, line 182: You can add mapping names (e.g., `mapping(uint128 pairId => uint32 fee)`) for clarity.

contracts/carbon/Strategies.sol
```diff
@@ -178,14 +178,22 @@ abstract contract Strategies is Initializable {
...
 
     /**
      * @dev triggered when the network fee is updated
      */
     event TradingFeePPMUpdated(uint32 prevFeePPM, uint32 newFeePPM);
 
+    /**
+     * @dev triggered when the trading fee override for a given pair is updated
+     */
+    event TradingFeePPMOverridesUpdated(uint128 pairId, uint32 prevFeePPM, uint32 newFeePPM);
```
lbeder, line 195: Perhaps `pairId` should be indexed?

contracts/carbon/Strategies.sol
```diff
@@ -514,17 +522,23 @@ abstract contract Strategies is Initializable {
     /**
      * @dev calculates the required amount plus fee
      */
-    function _addFee(uint128 amount) private view returns (uint128) {
+    function _addFee(uint128 amount, uint128 pairId) private view returns (uint128) {
+        // override protocol-wide trading fee with custom one if it's set for the pair
+        uint32 tradingFeePPMOverrides = _tradingFeePPMOverrides[pairId];
```
lbeder, line 527: Thinking out loud: what about encoding (e.g., in a bit) in the pair ID that it has a custom fee? This way, we would know to avoid a storage read for every pair that does not.

contracts/carbon/Strategies.sol
```diff
@@ -514,17 +522,23 @@ abstract contract Strategies is Initializable {
...
     }
 
     /**
      * @dev calculates the expected amount minus fee
      */
-    function _subtractFee(uint128 amount) private view returns (uint128) {
+    function _subtractFee(uint128 amount, uint128 pairId) private view returns (uint128) {
+        // override protocol-wide trading fee with custom one if it's set for the pair
+        uint32 tradingFeePPMOverrides = _tradingFeePPMOverrides[pairId];
+        uint32 tradingFeePPM = tradingFeePPMOverrides == 0 ? _tradingFeePPM : tradingFeePPMOverrides;
```
barakman, line 539: Suggesting to reuse this code in a single (private) helper function.
Especially in light of the fact that bytecode size is close to the limit.

contracts/carbon/interfaces/ICarbonController.sol
```diff
@@ -20,6 +20,11 @@ interface ICarbonController is IUpgradeable {
      */
     function tradingFeePPM() external view returns (uint32);
 
+    /**
+     * @dev returns the overriden trading fee for a given pair (in units of PPM)
+     */
+    function tradingFeePPMOverrides(uint128 pairId) external view returns (uint32);
```
barakman, line 26: I think that the main problem here (and in several other places) is that you're exposing what used to be an internal implementation detail up until now - `pairId` - as a new part in the API.

Consider keep using a pair of tokens as the formal identifier (in all interface functions, events, etc).

barakman (changes requested):

contracts/carbon/Strategies.sol
barakman: Or even a pair of tokens instead, since `pairId` is an internal implementation detail.

contracts/carbon/interfaces/ICarbonController.sol
ivanzhelyazkov: Pair ID is an identifier for the token pair, correct ?
Each token pair is associated with two token addresses. In a mapping I can't use two tokens, so it makes sense to have the mapping use pair id as a key and return the fee for that pair. 
It makes sense to keep the interface consistent with the view function and the event, since the mapping uses the pair id.

Keep in mind if that is changed, it will cost a SLOAD for the pair for both the view function and the event (Have to look up the tokens by pair id first).

contracts/carbon/CarbonController.sol
ivanzhelyazkov: Changed to `customTradingFeePPM`

contracts/carbon/Strategies.sol
ivanzhelyazkov: Added

contracts/carbon/Strategies.sol
ivanzhelyazkov: Added indexing

contracts/carbon/Strategies.sol
ivanzhelyazkov: I think that's a bit too complicated an optimization to save a storage read.
Regardless, I'm open for a discussion if we'd want to change it.

contracts/carbon/Strategies.sol
ivanzhelyazkov: Extracted repeating logic in a helper function.

contracts/carbon/interfaces/ICarbonController.sol
barakman: `Pair ID is an identifier for the token pair, correct ?` - an internal implementation detail.
Externally, the pair itself is used as the identifier.
I did not mean to complicate the implementation, only to pass the pair of tokens as part of the interface.
Inside the function, you can easily extract the pair ID.

ivanzhelyazkov: I would prefer to keep this interface function for consistency reasons.
What I can do is add an override to this function which accepts the token pair and returns the custom fee.
Open for discussion.

ivanzhelyazkov: I get what you mean, one step will be skipped (querying `pair(token0, token1)` in order to get the pair id when calling the first two functions).
For the event, it definitely makes sense to change since there is no public `getPairByPairId` function - so the event won't really help in tracing which pair has been updated unless we use an indexing service to track and store the pairs created.

I think this is worth implementing.

yudilevi (approved):

contracts/carbon/interfaces/ICarbonController.sol
ivanzhelyazkov: Replaced pair id with a token pair for each the two functions and event.

contracts/carbon/CarbonController.sol
```diff
@@ -133,6 +141,22 @@ contract CarbonController is
...
+     * @dev sets the custom trading fee for a given pair (in units of PPM)
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
+     */
+    function setCustomTradingFeePPM(
+        Token token0,
+        Token token1,
+        uint32 newCustomTradingFeePPM
```
barakman, line 154: I think that elsewhere in the code, we're using `Token[2] memory tokens` as function input.

contracts/carbon/Strategies.sol
```diff
@@ -514,17 +522,27 @@ abstract contract Strategies is Initializable {
...
+        uint32 tradingFeePPM = _getTradingFeePPM(pairId);
         // multiply the input amount by `1 - fee`
-        return MathEx.mulDivF(amount, PPM_RESOLUTION - _tradingFeePPM, PPM_RESOLUTION).toUint128();
+        return MathEx.mulDivF(amount, PPM_RESOLUTION - tradingFeePPM, PPM_RESOLUTION).toUint128();
+    }
+
+    /**
+     * @dev get the custom trading fee ppm for a given pair (returns default trading fee if not set for pair)
+     */
+    function _getTradingFeePPM(uint128 pairId) private view returns (uint32) {
```
barakman, line 543: Minor: I think that we're trying to avoid "get" as part of function naming (although I don't have a better suggestion here).

contracts/carbon/Strategies.sol
```diff
@@ -178,14 +178,22 @@ abstract contract Strategies is Initializable {
...
 
+    // mapping between a pair id to its custom trading fee (in units of PPM)
+    mapping(uint128 pairId => uint32 fee) internal _customTradingFeePPM;
+
     // upgrade forward-compatibility storage gap
-    uint256[MAX_GAP - 4] private __gap;
+    uint256[MAX_GAP - 5] private __gap;
 
     /**
      * @dev triggered when the network fee is updated
```
barakman, line 188: Minor: "network" is probably a copy/paste mistake from previous code.

barakman (changes requested):

contracts/carbon/CarbonController.sol
ivanzhelyazkov: We use both :)
Reference: check out `createStrategy` and `_createStrategy`.
The public function uses `token0` and `token1`.
Anyway, this is not a big issue imo.

contracts/carbon/Strategies.sol
ivanzhelyazkov: This is not a part of this PR.

yudilevi (approved):

lbeder (approved):

barakman (approved):

deploy/scripts/0007-CarbonController-upgrade.ts
```diff
@@ -0,0 +1,23 @@
...
+const func: DeployFunction = async ({ getNamedAccounts }: HardhatRuntimeEnvironment) => {
+    const { deployer } = await getNamedAccounts();
+    const voucher = await DeployedContracts.Voucher.deployed();
+
+    const carbonController = await DeployedContracts.CarbonController.deployed();
+    await upgradeProxy({
+        name: InstanceName.CarbonController,
+        from: deployer,
+        args: [voucher.address, carbonController.address],
+        initImpl: true
```
lbeder, line 17: This can be a good opportunity to remove `initImpl` and initialize the implementation construct in its contstructor instead (discussed with @yudilevi a few weeks ago)

lbeder (approved):

deploy/scripts/0007-CarbonController-upgrade.ts
yudilevi: Yup, planned as a separate PR

barakman (approved):

barakman (approved):
I suggest getting rid of the original tradingFeePPM function, because it will:
1. Cause confusion regarding which function should be called
2. Keep the already large byte-code larger than really needed
3. Allow anyone who is already using the contract, to overlook this change and keep calling the old function

yudilevi (approved):

lbeder (approved):

PR #112, opened 2023-08-12
Reduce the gas cost of trading on constant-rate orders

barakman:
Reduce the gas cost of trading on constant-rate orders, by updating `z` only when `A != 0`.

SDK:
1. When creating/updating a constant-rate order (`A = 0`), can pass `z = 0` in order to reduce user's gas cost
2. When updating a constant-rate order to dynamic-rate, should not rely on current `z` in order to determine new `z`

This PR is an alternative for https://github.com/bancorprotocol/carbon-sdk/pull/25.

### DIFF
--- a/contracts/carbon/Strategies.sol
+++ b/contracts/carbon/Strategies.sol
@@ -435,7 +435,7 @@ abstract contract Strategies is Initializable {
             }
 
             sourceOrder.y += sourceAmount;
-            if (sourceOrder.z < sourceOrder.y) {
+            if (sourceOrder.z < sourceOrder.y && sourceOrder.A != 0) {
                 sourceOrder.z = sourceOrder.y;
             }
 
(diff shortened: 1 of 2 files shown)


contracts/carbon/Strategies.sol
```diff
@@ -435,7 +435,7 @@ abstract contract Strategies is Initializable {
             }
 
             sourceOrder.y += sourceAmount;
-            if (sourceOrder.z < sourceOrder.y) {
+            if (sourceOrder.z < sourceOrder.y && sourceOrder.A != 0) {
```
lbeder, line 438: How much gas does it save? This seems like a non-trivial change which will require a careful review.

barakman: I agree.

It can save gas for the specific case of a trader selling his/her liquidity to a constant-rate order.
Potentially, 5k gas on average, for updating a slot from non-zero to non-zero (increasing the `z` value).

This will be particularly noticeable when trading with a single-order strategy, since the trader will be buying from that single order, hence - selling to the other order, which is a constant-rate order (of rate 0).

And it will actually happen on every trade on that irreversible strategy, until it is depleted (i.e., trader liquidity added to the invalid order on every trade, and the `z` value of the invalid order is updated on every trade).

barakman:
This contract supports range orders in general, with limit orders being a special case of them.
The new contract will support gradient orders, with limit orders being a special case of them.
The optimization suggested here can be implemented in the new contract, where strategies do not maintain a 'z' attribute.
This will allow creating limit-order strategies supporting cheaper trading.
Regardless of all of the above, the current contract is already immutable at this point, making this PR irrelevant.
Same goes for the corresponding PR in the SDK repository, at https://github.com/bancorprotocol/carbon-sdk/pull/25.

PR #115, opened 2023-08-24
Carbon POL

ivanzhelyazkov:
* Add Carbon POL contract implementation - swaps the Protocol-owned liquidity to ETH using Uni V3
* Add Carbon POL tests
* Add Carbon POL deployment scripts

### DIFF
--- a/contracts/pol/CarbonPOL.sol
+++ b/contracts/pol/CarbonPOL.sol
@@ -137,24 +137,36 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
     }
 
     /**
-     * @inheritdoc ICarbonPOL
+     * @dev sets the rewards ppm
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
      */
     function setRewardsPPM(uint32 newRewardsPPM) external onlyAdmin validFee(newRewardsPPM) {
         _setRewardsPPM(newRewardsPPM);
     }
 
     /**
-     * @inheritdoc ICarbonPOL
+     * @dev sets the max slippage ppm
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
      */
     function setMaxSlippagePPM(uint32 newMaxSlippagePPM) external onlyAdmin validFee(newMaxSlippagePPM) {
         _setMaxSlippagePPM(newMaxSlippagePPM);
     }
 
     /**
-     * @inheritdoc ICarbonPOL
+     * @dev sets the max tradeable ppb
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
      */
-    function setMaxTradeablePPB(uint32 newmaxTradeablePPB) external onlyAdmin validValue(newmaxTradeablePPB) {
-        _setMaxTradeablePPB(newmaxTradeablePPB);
+    function setMaxTradeablePPB(uint32 newMaxTradeablePPB) external onlyAdmin validPPBValue(newMaxTradeablePPB) {
+        _setMaxTradeablePPB(newMaxTradeablePPB);
     }
 
     /**
@@ -205,8 +217,8 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
         // cache rewardsPPM to save gas
         uint32 rewardsPercentage = _rewardsPPM;
         // initialize trade and reward amount arrays
-        uint256[] memory tradeAmounts = new uint256[](tokens.length);
-        uint256[] memory rewardAmounts = new uint256[](tokens.length);
+        uint256[] memory tradeAmounts = new uint256[](len);
+        uint256[] memory rewardAmounts = new uint256[](len);
         uint256 totalWethAmountOut;
 
         // trade tokens for wETH using Uniswap V3
--- a/contracts/pol/CarbonPOL.sol
+++ b/contracts/pol/CarbonPOL.sol
@@ -26,7 +26,7 @@ interface IWETH {
 }
 
 /**
- * @dev CarbonPOL contract
+ * @notice CarbonPOL contract
  */
 contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils {
     using Address for address payable;
@@ -137,7 +137,7 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
     }
 
     /**
-     * @dev sets the rewards ppm
+     * @notice sets the rewards ppm
      *
      * requirements:
      *
@@ -148,7 +148,7 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
     }
 
     /**
-     * @dev sets the max slippage ppm
+     * @notice sets the max slippage ppm
      *
      * requirements:
      *
@@ -159,7 +159,7 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
     }
 
     /**
-     * @dev sets the max tradeable ppb
+     * @notice sets the max tradeable ppb
      *
      * requirements:
      *
@@ -242,6 +242,9 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
         emit TokensTraded(msg.sender, tokens, poolFees, tradeAmounts, rewardAmounts, totalWethAmountOut);
     }
 
+    /**
+     * @dev trades token for wETH on Uniswap V3, and transfers token reward to user
+     */
     function _convertTokenAndAllocateRewards(
         Token token,
         uint24 poolFee,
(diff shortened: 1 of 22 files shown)


contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,415 @@
...
+        uint256 balance = token.balanceOf(address(this));
+        return _maxTradeableAmount(token, balance);
+    }
+
+    /**
+     * @inheritdoc ICarbonPOL
+     */
+    function tradeTokens(
+        Token[] calldata tokens,
+        uint24[] calldata poolFees
```
barakman, line 201: Consider an array of objects instead, where each object is of type:
```
struct TradeToken {
    Token token,
    uint24 poolFee
}
```

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,415 @@
...
+    /**
+     * @dev helper function to calculate the max tradeable amount given a token and the trade amount
+     * @dev max tradeable amount increases with each block and gets reset at each trade for a token
+     */
+    function _maxTradeableAmount(Token token, uint256 tradeAmount) private view returns (uint256) {
+        uint256 lastBlockTraded = _lastTradedBlock[token];
+        // if first trade for a token, check how many blocks have been mined since deployment
+        uint256 lastBlock = lastBlockTraded == 0 ? _deploymentBlock : lastBlockTraded;
+        uint256 blocksMined = block.number - lastBlock;
+        uint256 maxPPBTradeable = _maxTradeablePPBPerBlock * blocksMined;
```
barakman, line 317: Why not rely on timestamp, and then the contract behavior does not depend on the the current block-mining-time?
If you choose to accept this suggestion, then you'll also need to change the initialization:
```
// 1% tradeable per day (equal to 1389 ppb per block, given 7200 blocks per day)
_setMaxTradeablePPB(1389);
```
And for 1%, just use 10000000 (`PPB_RESOLUTION / 100`).

barakman (changes requested):

contracts/pol/interfaces/ICarbonPOL.sol
```diff
@@ -0,0 +1,109 @@
...
+     */
+    function rewardsPPM() external view returns (uint32);
+
+    /**
+     * @dev returns the max slippage ppm
+     */
+    function maxSlippagePPM() external view returns (uint32);
+
+    /**
+     * @dev returns the max tradeable ppb per block
```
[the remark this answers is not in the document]
ivanzhelyazkov: It's per block. Set initially to 1389 points per billion, because we discussed having the max tradeable amount increasing by 1% of the token balance every 24 hours.
Since the merge, blocks are mined every 12 seconds precisely, so 5 per minute. That means, we have exactly 5 * 60 * 24 = 7,200 blocks every 24 hours. To get the max tradeable % increase per block, we need to divide 1% by 7,200 blocks = 0.000001389% per block.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,415 @@
...
+    uint256 private immutable _deploymentBlock;
+
+    // rewards percentage
+    uint32 private _rewardsPPM;
+
+    // max slippage percentage
+    uint32 private _maxSlippagePPM;
+
+    // max tradeable ppb (parts per billion) per block
+    uint32 private _maxTradeablePPBPerBlock;
```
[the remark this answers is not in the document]
ivanzhelyazkov: It's per block. On the fence for the name change - if there's someone else who prefers this convention, will change.

contracts/utility/Utils.sol
```diff
@@ -59,4 +61,32 @@ abstract contract Utils {
...
+
+    // ensures that the value is valid
+    modifier validValue(uint32 value) {
+        _validValue(value);
+
+        _;
+    }
+
+    // error message binary size optimization
+    function _validValue(uint32 value) internal pure {
```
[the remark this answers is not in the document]
ivanzhelyazkov: validPPBValue ?
Open to suggestions here

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: A lot of implementation changes required for this with little to no benefit.
Blocks get mined at even intervals, so we can rely on them for calculating the increase for any duration of time.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Agreed on the approach, seems better, but there's an open discussion about the poolFees currently, so let's decide later.

contracts/pol/interfaces/ICarbonPOL.sol
```diff
@@ -0,0 +1,109 @@
...
+ * @dev CarbonPOL interface
+ */
+interface ICarbonPOL is IUpgradeable {
+    error DuplicateToken();
+    error InvalidToken();
+    error InvalidTokenLength();
+    error TokenPoolFeesLengthMismatch();
+
+    /**
+     * @dev triggered after a successful trade is executed
```
lbeder, line 17: Nitpick: Eventually, we should replace theses @devs with @notice when applicable

contracts/pol/interfaces/ICarbonPOL.sol
```diff
@@ -0,0 +1,109 @@
...
+    function setMaxSlippagePPM(uint32 newMaxSlippagePPM) external;
+
+    /**
+     * @dev sets the max tradeable ppb
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
+     */
+    function setMaxTradeablePPB(uint32 newMaxTradeablePPB) external;
```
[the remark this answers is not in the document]
ivanzhelyazkov: Removed admin functions from interface

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,427 @@
...
+        }
+
+        // convert token to wETH using Uniswap V3
+        wethReceived = _convertTokenToETH(token, poolFee, tradeAmount);
+        // transfer reward to caller
+        // @note here token can't be native - we revert in `validateTokens` if it is
+        token.safeTransfer(msg.sender, rewardAmount);
+    }
+
+    function _convertTokenToETH(Token token, uint24 poolFee, uint256 tradeAmount) private returns (uint256 amountOut) {
```
lbeder, line 276: Nitpick: missing description

contracts/pol/CarbonPOL.sol
lbeder: Also don't forget that timestamps aren't necessary monotonic

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,427 @@
...
+     */
+    function _maxTradeableAmount(Token token, uint256 tradeAmount) private view returns (uint256) {
+        uint256 lastBlockTraded = _lastTradedBlock[token];
+        // if first trade for a token, check how many blocks have been mined since deployment
+        uint256 lastBlock = lastBlockTraded == 0 ? _deploymentBlock : lastBlockTraded;
+        uint256 blocksMined = block.number - lastBlock;
+        uint256 maxPPBTradeable = _maxTradeablePPBPerBlock * blocksMined;
+        // max tradeable is capped at 100% of tradeAmount
+        if (maxPPBTradeable > PPB_RESOLUTION) {
+            maxPPBTradeable = PPB_RESOLUTION;
```
lbeder, line 332: Nitpick: can use Math.min here

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,427 @@
...
+     * @dev set allowance to Uniswap V3 to the max amount if it's less than the input amount
+     */
+    function _setAllowance(Token token, uint256 inputAmount) private {
+        if (token.isNative()) {
+            return;
+        }
+        uint256 allowance = token.allowance(address(this), address(_uniswapV3Router));
+        if (allowance < inputAmount) {
+            // increase allowance to the max amount if allowance < inputAmount
+            token.safeIncreaseAllowance(address(_uniswapV3Router), type(uint256).max - allowance);
```
lbeder, line 393: If you'll upgrade OZ to 4.9.3, this will use the newer (and more gas efficient) `forceApprove`

lbeder (commented):
Finished my initial overview, but I'd still need to perform a full one later, so FYI

contracts/pol/CarbonPOL.sol
barakman: And if you wait for the version currently on their master branch (going to the next release), they have applied some additional improvements in the increase/decrease allowance functions.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Seems a bit excessive to import an entire lib for this line.

barakman: The entire lib's code is not linked into the contract.
Only the functions which you actually use, are linked into the contract.

ivanzhelyazkov: I am aware. Even though, it's still one line of code.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Estimations on the gas cost savings ?

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: added description

contracts/pol/interfaces/ICarbonPOL.sol
ivanzhelyazkov: replaced

contracts/pol/CarbonPOL.sol
lbeder: Basically, it should be almost the same. The main difference is that the newer approach has also a fallback to try approve(0) if the original approve fails (e.g., for USDT)

PR #116, opened 2023-09-07
Carbon POL

ivanzhelyazkov:
* Add Carbon POL contract
* Add Carbon POL tests
* Add Carbon POL deployment scripts


contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        // get initial price as set by enableTrading
+        price = _initialPrice[token];
+        // calculate the actual price by multiplying the eth amount by the factor
+        price.ethAmount *= _marketPriceMultiply;
+        // get time elapsed since trading was enabled
+        uint32 timeElapsed = uint32(block.timestamp) - _tradingStartTimes[token];
+        // get the current price by adjusting the eth amount with the exp decay formula
+        price.ethAmount = uint128(ExpDecayMath.calcExpDecay(price.ethAmount, timeElapsed, _priceDecayHalfLife));
+        // return the price
+        return price;
```
barakman, line 217: This solution can be greatly simplified, as well as improved for accuracy.

Function `calcExpDecay` performs a division, which is redundant in this case, since the `Price` object already represents `ethAmount / tokenAmount`, so there is no need for an additional division.

So in short, you can just use whatever function `exp2` returns, which means that you don't even need to include function `calcExpDecay` as part of the project, and you don't need to worry about any additional testing of your changes in that function (since in contrast to it, function `exp2` remains unchanged).

And as an additional bonus, you avoid the loss of accuracy which occurs when function `calcExpDecay` calls function `mulDivF`.

In short, here is the code:
```
function tokenPrice(Token token) public view returns (Price memory) {
    Price memory initialPrice = _initialPrice[token];

    Fraction memory timeRatio = Fraction({
        n: block.timestamp - _tradingStartTimes[token],
        d: _priceDecayHalfLife }
    );

    Fraction memory decayFactor = MathEx.exp2(timeRatio);

    return Price({
        ethAmount: initialPrice.ethAmount * decayFactor.d * _marketPriceMultiply,
        tokenAmount: initialPrice.tokenAmount * decayFactor.n
    });
}
```

Lastly, since we already have a `Fraction` object, you might want to consider it instead of the `Price` object (though the `Fraction` object is less verbose than the `Price` object, so just FYI to consider).

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /**
+     * @notice enable trading for a token and set the initial price
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
+     */
+    function enableTrading(Token token, Price memory price) external onlyAdmin validPrice(price) {
```
lbeder, line 135: This method can be called multiple times and cause some unexpected results (e.g., it can reset or even decrease the price significantly from where the system expects the "initial price" to be; it can move its trading start time to the past, etc.). 

I'd suggest (not necessarily mutual exclusive):
1. Explicitly prevent this method from being called twice
2. Add another method that updates the price or the starting time explicitly 
3. Similar to (2), you may consider adding a method that disables and resets trading for a token (e.g., as a response to a new hack/exploit, newly discovered bug in the token, etc.). This emergency action should be vetted by the DAO ahead of time.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+     * @inheritdoc ICarbonPOL
+     */
+    function expectedTradeReturn(Token token, uint128 ethAmount) external view returns (uint128 tokenAmount) {
+        // return 0 if trading is not enabled for token
+        if (!_tradingEnabled(token)) {
+            return 0;
+        }
+        Price memory currentPrice = tokenPrice(token);
+        // revert if ethAmount in price is equal to 0
+        if (currentPrice.ethAmount == 0) {
```
lbeder, line 172: To be on the safe side and future proof, I'd suggest using `_validPrice` instead.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        return _tradingEnabled(token);
+    }
+
+    /**
+     * @inheritdoc ICarbonPOL
+     */
+    function expectedTradeReturn(Token token, uint128 ethAmount) external view returns (uint128 tokenAmount) {
+        // return 0 if trading is not enabled for token
+        if (!_tradingEnabled(token)) {
+            return 0;
```
lbeder, line 168: I, personally, don't like mixing return declarations (`returns (uint128 tokenAmount)`) with explicit returns. It's very error-prone. Please consider refactoring 🙏

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        }
+    }
+
+    /**
+     * @inheritdoc ICarbonPOL
+     */
+    function expectedTradeInput(Token token, uint128 tokenAmount) public view returns (uint128 ethAmount) {
+        // return 0 if trading is not enabled for token
+        if (!_tradingEnabled(token)) {
+            return 0;
```
lbeder, line 189: Same here

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        // set market price multiplier to 2x
+        _setMarketPriceMultiply(2);
+        // set price decay half-life to 10 days
+        _setPriceDecayHalfLife(10 days);
+    }
+
+    /**
+     * @dev authorize the contract to receive the native token
+     */
+    receive() external payable {}
```
lbeder, line 77: Why is this needed? The contract shouldn't receive ETH explicitly (since it can't really offer it for trading) and traded ETH is receiving with an explicit `payable` function anyway.

contracts/pol/CarbonPOL.sol
lbeder: Similar to this note, it would be dangerous to allow modification of `_priceDecayHalfLife` for ongoing auctions (with some extra precautions)

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        return uint128(MathEx.mulDivF(currentPrice.ethAmount, tokenAmount, currentPrice.tokenAmount));
+    }
+
+    /**
+     * @inheritdoc ICarbonPOL
+     */
+    function tokenPrice(Token token) public view returns (Price memory price) {
+        // return 0 if trading hasn't been enabled for a token
+        if (!_tradingEnabled(token)) {
+            return Price({ tokenAmount: 0, ethAmount: 0 });
```
lbeder, line 206: Same here re: mixing

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+    function tradingEnabled(Token token) external view returns (bool) {
+        return _tradingEnabled(token);
+    }
+
+    /**
+     * @inheritdoc ICarbonPOL
+     */
+    function expectedTradeReturn(Token token, uint128 ethAmount) external view returns (uint128 tokenAmount) {
+        // return 0 if trading is not enabled for token
+        if (!_tradingEnabled(token)) {
```
lbeder, line 167: `_tradingEnabled(token)` repeats a few times. Perhaps it should be refactored to a modifier? Note that it's also used by `_validToken`, so in some cases it can be used instead.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        // check enough eth has been sent for the trade
+        if (msg.value < ethRequired) {
+            revert InsufficientNativeTokenSent();
+        }
+        // transfer the tokens to caller
+        token.safeTransfer(msg.sender, amount);
+
+        // refund any excess eth to caller
+        if (msg.value > ethRequired) {
+            payable(msg.sender).sendValue(msg.value - ethRequired);
```
lbeder, line 241: Super-duper nitpick: the subtraction can be `unchecked`

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /**
+     * @dev return true if trading is enabled for token
+     */
+    function _tradingEnabled(Token token) private view returns (bool) {
+        return _tradingStartTimes[token] != 0;
+    }
+
+    function uncheckedInc(uint256 i) private pure returns (uint256 j) {
```
lbeder, line 298: Nitpick: can be moved to something like `Utils`

lbeder: Nitpick #2: not being used by the contract

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+import { Token } from "../token/Token.sol";
+import { Utils } from "../utility/Utils.sol";
+import { MathEx } from "../utility/MathEx.sol";
+import { ExpDecayMath } from "../utility/ExpDecayMath.sol";
+import { MAX_GAP, PPM_RESOLUTION } from "../utility/Constants.sol";
+
+/**
+ * @notice CarbonPOL contract
+ */
+contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils {
```
lbeder, line 19: BTW, any reason why we have stopped using `onlyProxyDelegate` in this (and other) contracts?

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        // refund any excess eth to caller
+        if (msg.value > ethRequired) {
+            payable(msg.sender).sendValue(msg.value - ethRequired);
+        }
+
+        // emit event
+        emit TokenTraded(msg.sender, token, amount, ethRequired);
+    }
+
+    function _setMarketPriceMultiply(uint32 newMarketPriceMultiply) private {
```
lbeder, line 248: Nitpick: no NATSPEC on this and other functions

lbeder (changes requested):

contracts/pol/CarbonPOL.sol
barakman: We seem to think [alike](https://github.com/bancorprotocol/carbon-contracts/pull/116#discussion_r1318809594).

contracts/pol/CarbonPOL.sol
barakman: We seem to think [alike](https://github.com/bancorprotocol/carbon-contracts/pull/116#discussion_r1318890099) and [alike](https://github.com/bancorprotocol/carbon-contracts/pull/116#discussion_r1318893381).

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Prone to overflow - decayFactor.d is actually 2 ^ _priceDecayHalfLife.
So for any _priceDecayHalfLife > 256 and initialPrice.ethAmount >= 1 the value overflows.
Same goes for tokenAmount.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Do you have a suggestion for this ?
I want to make it explicit that the function is returning tokenAmount for clarity, same with expectedTradeInput returning ethAmount.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Prefer to leave it as regular subtraction - doesn't save that much gas and makes the code less readable.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    /**
+     * @inheritdoc ICarbonPOL
+     */
+    function trade(
+        Token token,
+        uint128 amount
+    ) external payable nonReentrant validToken(token) greaterThanZero(amount) {
+        uint128 ethRequired = expectedTradeInput(token, amount);
```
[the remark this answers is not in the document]
ivanzhelyazkov: Different types of checks - we get a more accurate error message (TradingDisabled) when user attempts to trade with a token for which trading is disabled with this approach at an extra 100 gas cost.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+ * @notice CarbonPOL contract
+ */
+contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils {
+    using Address for address payable;
+
+    // initial starting price multiplier for the dutch auction
+    uint32 private _marketPriceMultiply;
+
+    // time until the price gets back to market price
+    uint32 private _priceDecayHalfLife;
```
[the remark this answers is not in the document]
ivanzhelyazkov: True, but doesn't align with the product design. Both should be settable by an admin, not updated via an upgrade.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: It's used in functions returning different values, so I don't think it makes sense to refactor it as a modifier.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+
+    // token to trading start time mapping
+    mapping(Token token => uint32 tradingStartTime) private _tradingStartTimes;
```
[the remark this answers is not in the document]
ivanzhelyazkov: There's no gas-related advantage, but it does enforce the types of values these variables should hold.
Also was agreed to in the design doc beforehand.
There are a few conversions related to the timestamp, but not too much.
Using uint128 for the Price saves around 50% of gas costs when admin is setting the token prices.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+
+    // token to initial price mapping
+    mapping(Token token => Price initialPrice) private _initialPrice;
```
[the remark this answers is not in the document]
ivanzhelyazkov: Price as part of the interface can't be imported explicitly. 
Only way is to define it outside of the interface and import it, but I tend to prefer defining it as a part of the interface.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Need to discuss pros and cons of each approach, agree in general that it's not a good idea to call it more than once for a token with a few exceptions.
Would consider setting price only once, while allowing for enabling / disabling trading for a token multiple times.
Need some more input here.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        Price memory currentPrice = tokenPrice(token);
+        // revert if ethAmount in price is equal to 0
+        if (currentPrice.ethAmount == 0) {
+            revert InvalidPrice();
+        }
+        // multiply the token amount by the eth amount / total eth amount ratio to get the actual tokens received
+        tokenAmount = uint128(MathEx.mulDivF(currentPrice.tokenAmount, ethAmount, currentPrice.ethAmount));
+        // revert if not enough token balance
+        if (tokenAmount > uint128(token.balanceOf(address(this)))) {
+            revert InsufficientTokenBalance();
```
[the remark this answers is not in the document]
ivanzhelyazkov: Agreed, but was discussed beforehand to revert in this case for both functions. 
Doesn't make sense to return a valid price up to an amount and suddenly 0. And a bit misleading to return amount if there are not enough tokens.
Might consider not reverting at all in this case, up to discussion.

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+     * @inheritdoc ICarbonPOL
+     */
+    function expectedTradeInput(Token token, uint128 tokenAmount) public view returns (uint128 ethAmount) {
+        // return 0 if trading is not enabled for token
+        if (!_tradingEnabled(token)) {
+            return 0;
+        }
+        // revert if not enough token balance for trade
+        if (tokenAmount > token.balanceOf(address(this))) {
+            revert InsufficientTokenBalance();
```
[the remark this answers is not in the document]
ivanzhelyazkov: Same as above

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Added

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Removed

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Implemented suggestion

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        if (price.tokenAmount == 0 || price.ethAmount == 0) {
+            revert InvalidPrice();
+        }
+    }
+
+    /**
+     * @dev return true if trading is enabled for token
+     */
+    function _tradingEnabled(Token token) private view returns (bool) {
+        return _tradingStartTimes[token] != 0;
```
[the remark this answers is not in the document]
ivanzhelyazkov: Implemented where needed.

contracts/helpers/TestExpDecayMath.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity 0.8.19;
+
+import { ExpDecayMath } from "../utility/ExpDecayMath.sol";
+
+contract TestExpDecayMath {
+    function calcExpDecay(uint256 ethAmount, uint32 timeElapsed, uint32 halfLife) external pure returns (uint256) {
```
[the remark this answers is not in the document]
ivanzhelyazkov: Updated

contracts/pol/interfaces/ICarbonPOL.sol
```diff
@@ -0,0 +1,88 @@
...
+     */
+    function expectedTradeInput(Token token, uint128 tokenAmount) external view returns (uint128 ethAmount);
+
+    /**
+     * @notice returns the current token price
+     */
+    function tokenPrice(Token token) external view returns (Price memory price);
+    
+    /**
+     * @notice trades *amount* of token for ETH based on the current token price (trade by source amount)
```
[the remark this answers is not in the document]
ivanzhelyazkov: Updated

contracts/pol/CarbonPOL.sol
```diff
@@ -0,0 +1,303 @@
...
+        }
+        Price memory currentPrice = tokenPrice(token);
+        // multiply the eth amount by the token amount / total token amount ratio to get the actual eth to send
+        return uint128(MathEx.mulDivF(currentPrice.ethAmount, tokenAmount, currentPrice.tokenAmount));
+    }
+
+    /**
+     * @inheritdoc ICarbonPOL
+     */
+    function tokenPrice(Token token) public view returns (Price memory price) {
```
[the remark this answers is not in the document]
ivanzhelyazkov: Updated

contracts/utility/ExpDecayMath.sol
```diff
@@ -0,0 +1,23 @@
...
+
+import { Fraction } from "./Fraction.sol";
+import { MathEx } from "./MathEx.sol";
+
+/**
+ * @dev This library supports the calculation of exponential price decay
+ */
+library ExpDecayMath {
+    /**
+     * @dev returns the amount of eth required for a token after a given time period since trading has been enabled
```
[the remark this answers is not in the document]
ivanzhelyazkov: Updated

contracts/utility/ExpDecayMath.sol
```diff
@@ -0,0 +1,23 @@
...
+     *
+     * the returned value is calculated as `ethAmount * (1 / 2 ^ (timeElapsed / halfLife))`
```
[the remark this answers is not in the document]
ivanzhelyazkov: Updated

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: I think there was an extra step necessary (upgrade) to set the proxy address for the contract, making it automatically start with version 2 to function normally.

contracts/pol/CarbonPOL.sol
barakman: We're not calculating `2 ^ halfLife`, we're calculating `2 ^ timeElapsed / halfLife`.

And the value of `timeElapsed / halfLife` cannot exceed ~23 (16 / ln 2) regardless, due to a technical limitation in function `exp2`.

If anything, the overflow here would occur because each component in the `Fraction` returned by `exp2` is scaled up by `2 ^ 127`.

And you can handle that by declaring the `Price` components as `uint256` instead of `uint128` (in case they aren't already declared as such).

contracts/pol/CarbonPOL.sol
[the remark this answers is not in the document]
barakman: Got it. So regarding the `uint128` - if possible, we should keep it contained only for the actual storage variables, i.e., all local variables and function input arguments should be `uint256`, which would help preventing overflows in intermediate computations.

For example (hypothetical), the actual price of a token is `uint128`, but this price is represented as a fraction of two components which are `uint256` each.

contracts/pol/CarbonPOL.sol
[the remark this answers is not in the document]
barakman: This gas saving is designated for admin and not for actual users; I believe that we should ditch it if it causes any issues (and it does, as far as I understand - if `tokenAmount` and `ethAmount` are 256 bits each, then we can calculate them accurately without the risk of overflow).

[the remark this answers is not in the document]
ivanzhelyazkov: Doesn’t have any risk, no v3 tokens are over 2^128 total supply, nor eth.
It’s used by multiple different exchanges / protocols.

contracts/pol/CarbonPOL.sol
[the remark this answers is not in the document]
barakman: Are you sure?

`validToken` calls `_validToken`, which calls `_tradingEnabled`, which reads `_tradingStartTimes`.

`expectedTradeInput` calls `_tradingEnabled`, which reads `_tradingStartTimes`.

So we end up reading `_tradingStartTimes` from storage twice in the same transaction.

I mean, the caller might be able to reduce that cost by using the `accessList` feature of the EVM, but do we want the code rely on that?

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: I tried this approach and it reverted with overflow.
I already implemented the other approach and wrote the tests for it.

contracts/pol/CarbonPOL.sol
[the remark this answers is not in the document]
ivanzhelyazkov: We read it twice, which means one cold read + one warm read. Warm read costs 100 gas.

contracts/pol/CarbonPOL.sol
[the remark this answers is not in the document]
ivanzhelyazkov: Added SafeCast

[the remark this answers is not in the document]
lbeder: Some projects are shifting to use `uint64` for timestamps instead, but on and all, it's insignificant in this context

contracts/pol/CarbonPOL.sol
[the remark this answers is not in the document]
lbeder: But changing them during an upgrade is equivalent to the requirement (just differs in implementation details and gas costs)

contracts/pol/CarbonPOL.sol
lbeder: I'm saying more than that: not only that it's not a great idea - it's a super risky and dangerous idea 😅. I like your suggestion, though. @yudilevi?

contracts/pol/CarbonPOL.sol
lbeder: I understand your rationale, but my note is that it's differs from the current style and practices used by the team (which is generally don't define explicit return variables, as they are prone to error, unless it's a solution for the notorious "Stack too Deep" issue)

contracts/pol/CarbonPOL.sol
lbeder: I'm not sure how return values are relevant for the discussion here. Using it as a "tradingEnabled" modifier will help DRYing out the code a bit.

contracts/pol/CarbonPOL.sol
lbeder: 👍

contracts/pol/CarbonPOL.sol
lbeder: What do you mean? `onlyProxyDelegate` is only used for IP protection.

contracts/pol/CarbonPOL.sol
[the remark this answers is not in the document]
ivanzhelyazkov: Agreed, but the implementation details were also defined beforehand. Waiting on some more input here.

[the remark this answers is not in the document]
yudilevi: Not critical to be super optimized here, better off keeping it flexible to minimize code changes in this case

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: If I understand what you mean correctly, you are talking about refactoring `_tradingEnabled(token)` as a modifier.
If we do that, we'd need to define the code in the modifier as such to return a value, 0 in this case. This is not possible with a modifier.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Logic here is a bit confusing and will need further testing to confirm and verify it's working correctly.
Sticking for my current solution for now, waiting for further input.
Your original suggestion is prone to overflow if using uint128, and switching to uint256 means we disregard the gas savings. I don't see much benefit of implementing it at this stage.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Changed

contracts/pol/CarbonPOL.sol
yudilevi: this was only part of the CarbonController contract

contracts/pol/CarbonPOL.sol
yudilevi: it's an admin only function that the DAO might want to use in edge cases (in cases the TKN price drops significantly - e.g. 10 years to get to the market price) to fix the price

contracts/pol/interfaces/ICarbonPOL.sol
```diff
@@ -0,0 +1,88 @@
...
+     */
+    function expectedTradeReturn(Token token, uint128 ethAmount) external view returns (uint128 tokenAmount);
+
+    /**
+     * @notice returns the expected trade input (how much eth to send) given an token amount received
+     */
+    function expectedTradeInput(Token token, uint128 tokenAmount) external view returns (uint128 ethAmount);
+
+    /**
+     * @notice returns the current token price
```
yudilevi, line 80: specify the direction here ("tkn / eth")

test/forge/CarbonPOL.t.sol
```diff
@@ -0,0 +1,744 @@
...
+        ICarbonPOL.Price memory initialPrice = ICarbonPOL.Price({ ethAmount: ethAmount, tokenAmount: tokenAmount });
+        carbonPOL.enableTrading(token, initialPrice);
+
+        // set timestamp to 1
+        vm.warp(1);
+
+        ICarbonPOL.Price memory price = carbonPOL.tokenPrice(token);
+        uint128 expectedETHAmount = getExpectedETHAmount(ethAmount);
+
+        assertEq(price.ethAmount, expectedETHAmount);
```
yudilevi, line 253: this tests the contract vs itself which basically means that it doesn't really test anything - I recommend creating some hardcoded scenarios and testing the values against expected (hardcoded) results in those scenarios

yudilevi (commented):
missing enable trading script

barakman (approved):

contracts/pol/CarbonPOL.sol
lbeder: I see. Perhaps worth adding a suitable comment so that it doesn't appear as a mistake. Resolving in any case.

contracts/pol/CarbonPOL.sol
lbeder: Acknowledged

contracts/pol/CarbonPOL.sol
lbeder: Ah, gotcha. Resolving.

contracts/pol/CarbonPOL.sol
```diff
@@ -200,17 +198,19 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
     /**
      * @inheritdoc ICarbonPOL
      */
-    function tokenPrice(Token token) public view returns (Price memory price) {
+    function tokenPrice(Token token) public view returns (Price memory) {
+        // cache trading start time to save gas
+        uint32 tradingStartTime = _tradingStartTimes[token];
         // return 0 if trading hasn't been enabled for a token
-        if (!_tradingEnabled(token)) {
+        if (tradingStartTime == 0) {
```
lbeder, line 208: Super nitpick (feel free to ignore and resolve): since this duplicated the logic of testing if trading is enabled (once here and once in `_tradingEnabled`), you may consider moving the logic to a shared pure function that receives the trading start time.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Addressed the issue

test/forge/CarbonPOL.t.sol
ivanzhelyazkov: added test cases for different initial prices and timestamps

contracts/pol/interfaces/ICarbonPOL.sol
ivanzhelyazkov: added

yudilevi (approved):

lbeder (approved):

PR #117, opened 2023-09-12
carbon pol - add enable trading script

ivanzhelyazkov:
* Add enable trading script for carbon pol tokens
* Script enables trading by setting the token price based on coingecko market data


scripts/enableTrading.ts
```diff
@@ -0,0 +1,184 @@
...
+    const tokens: Record<string, TokenData> = {};
+
+    let symbol: string;
+    let decimals: number;
+    for (let i = 0; i < allTokens.length; i++) {
+        const token = allTokens[i];
+
+        const tokenContract = await Contracts.ERC20.attach(token, deployer);
+        symbol = await tokenContract.symbol();
+        decimals = await tokenContract.decimals();
```
lbeder, line 76: Note that this will fail for a few tokens (i.e., an exception will be thrown), but we can cross this bridge when we'd get there...

scripts/enableTrading.ts
```diff
@@ -0,0 +1,184 @@
...
+
+const MAX_PRECISION = 16;
+
+const main = async () => {
+    const { deployer } = await getNamedSigners();
+    const carbonPOL = await DeployedContracts.CarbonPOL.deployed();
+
+    const tokenAddressesEnv = process.env.TOKEN_ADDRESSES;
+    if (!tokenAddressesEnv) {
+        console.error(`no tokens passed in - pass in token addresses like so: TOKEN_ADDRESSES='0x...','0x...'`);
```
lbeder, line 30: Nitpick: Logger.error

scripts/enableTrading.ts
```diff
@@ -0,0 +1,184 @@
...
+        return;
+    }
+
+    // Remove single quotes and whitespace, then split by commas
+    const allTokens = tokenAddressesEnv
+        .replace(/'/g, '')
+        .split(',')
+        .map((address) => address.trim());
```
lbeder, line 38: Suggestion: store allTokens as `Set<string>` and normalize the addresses by calling ethers' `getAddress` (or lower/upper case the address). This way, you will automatically ignore duplicates.

lbeder (changes requested):

scripts/enableTrading.ts
yudilevi: added support for it

scripts/enableTrading.ts
yudilevi: removed that code altogether

scripts/enableTrading.ts
yudilevi: removed that code altogether

lbeder (approved):

lbeder (approved):

PR #119, opened 2023-10-11
Reentrancy tests

ivanzhelyazkov:
* Add reentrancy tests for CarbonController, CarbonVortex and CarbonPOL

### DIFF
--- a/contracts/helpers/TestReentrantToken.sol
+++ b/contracts/helpers/TestReentrantToken.sol
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.19;
+
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+import { ICarbonController } from "../carbon/interfaces/ICarbonController.sol";
+import { Order, TradeAction } from "../carbon/Strategies.sol";
+import { Token } from "../token/Token.sol";
+
+/**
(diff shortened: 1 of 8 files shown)


contracts/helpers/TestReentrantToken.sol
```diff
@@ -0,0 +1,120 @@
...
+        CREATE_STRATEGY,
+        UPDATE_STRATEGY,
+        DELETE_STRATEGY,
+        TRADE_BY_SOURCE_AMOUNT,
+        TRADE_BY_TARGET_AMOUNT,
+        WITHDRAW_FEES
+    }
+
+    // which function to reenter using transferFrom
+    uint8 private _reenterFunctionIndex;
```
lbeder, line 27: Can you define it as `ReenterFunctions`? This way you won't need the casts below.

ivanzhelyazkov: It's defined as uint8 so as to be able to pass it as a fuzzing parameter in the test testStrategyCreationRevertsIfReentrancyIsAttempted(uint8 reenterFunctionIndex) in Strategies.t.sol - so that we don't need to define several tests for it.

ivanzhelyazkov: fixed

lbeder (approved):

yudilevi (approved):

PR #120, opened 2023-11-15
POL BNT -> ETH conversion

ivanzhelyazkov:
* Add BNT -> ETH trades to POL contract using a dutch auction:
Based on this [Protocol Owned ETH Spending Proposition](https://snapshot.org/#/bancornetwork.eth/proposal/0x0aafb738f6b59612b5695ae9d09ed973882509df86a934729426e3acedc709e2)
* Add tests
* Add deployment scripts


contracts/pol/interfaces/ICarbonPOL.sol
```diff
@@ -50,28 +66,42 @@ interface ICarbonPOL is IUpgradeable {
...
+    function currentEthSaleAmount() external view returns (uint128);
+
     /**
      * @notice returns true if trading is enabled for token
      */
     function tradingEnabled(Token token) external view returns (bool);
 
     /**
      * @notice returns the expected trade output (tokens received) given an eth amount sent for a token
+     * @notice if token == ETH, return how much bnt will be sent given an eth amount received
```
lbeder, line 86: Nitpick: there is some inconsistency using bnt, eth, BNT, and ETH

contracts/pol/CarbonPOL.sol
```diff
@@ -33,13 +36,17 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
...
     // upgrade forward-compatibility storage gap
-    uint256[MAX_GAP - 3] private __gap;
+    uint256[MAX_GAP - 4] private __gap;
 
     /**
      * @dev used to initialize the implementation
      */
-    constructor() {
+    constructor(Token bntInit) {
+        _bnt = bntInit;
```
lbeder, line 49: Not super important, but I'd consider checking that `bntInit` (`initBnt`?) isn't 0x0

contracts/pol/CarbonPOL.sol
```diff
@@ -128,18 +137,47 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
...
      * requirements:
      *
      * - the caller must be the admin of the contract
+     * - can only enable trading for non-native tokens
      */
     function enableTrading(Token token, Price memory price) external onlyAdmin validPrice(price) {
+        if (token == NATIVE_TOKEN) {
+            revert InvalidToken();
+        }
         _tradingStartTimes[token] = uint32(block.timestamp);
```
lbeder, line 162: There weren't any ETH trades enabled already, right? If there were, then I'd consider explicitly "resetting" everything re: ETH during the upgrade

contracts/pol/CarbonPOL.sol
```diff
@@ -154,6 +192,20 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
...
+     * @inheritdoc ICarbonPOL
+     */
+    function ethSaleAmount() external view returns (uint128) {
+        return _ethSaleAmount.initial;
+    }
+
+    /**
+     * @inheritdoc ICarbonPOL
+     */
+    function currentEthSaleAmount() external view returns (uint128) {
```
lbeder, line 205: I'd consider returning the `EthSaleAmount memory` directly instead of splitting it into two functions

lbeder (commented):
I think this should've been simpler to implement in a separate CarbonPOL-like contract, while sharing some functionality with the original CarbonPOL contract via a library. It will prevent some of the side effects and "double meaning". 

Let's discuss and then I'll continue my review 🙏

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: I don't think so, but that's a good note - this is why this change was added - enabling ETH from this function messes up the new logic. Worth checking imo @yudilevi

ivanzhelyazkov:
> I think this should've been simpler to implement in a separate CarbonPOL-like contract, while sharing some functionality with the original CarbonPOL contract via a library. It will prevent some of the side effects and "double meaning".
> 
> Let's discuss and then I'll continue my review 🙏

I think that for the time being this should be sufficient - I realize there are some inconsistencies, this is why I wanted to split the logic in buy and sell - different prices, trade start times and trade functions.
But given that a new integration would be required and the technical time for that, and that we're only adding ETH->BNT trades, it's easier to keep the current interface with no changes (much simpler to integrate)  .

contracts/pol/interfaces/ICarbonPOL.sol
ivanzhelyazkov: Fixed

contracts/pol/CarbonPOL.sol
lbeder: IMHO, as someone who interacts with contracts all the time, both options are equivalvently user friendly. Every modern framework (wagmi, ethers, viem, etc.) handles input/ouput structs very easily and convinently.

ivanzhelyazkov: Changed - personally, I'm not a fan of the `carbonPOL.ethSaleAmount().initial` syntax.

contracts/pol/CarbonPOL.sol
yudilevi: not following you but no ETH -> BNT trading was ever enabled

contracts/pol/CarbonPOL.sol
```diff
@@ -164,17 +210,19 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
...
         _validPrice(currentPrice);
-        // multiply the token amount by the eth amount / total eth amount ratio to get the actual tokens received
-        uint128 tokenAmount = MathEx.mulDivF(currentPrice.tokenAmount, ethAmount, currentPrice.ethAmount).toUint128();
+        // calculate the trade return based on the current price and token
+        uint128 tradeReturn = MathEx
+            .mulDivF(currentPrice.targetAmount, tradeInput, currentPrice.sourceAmount)
+            .toUint128();
         // revert if not enough token balance
-        if (tokenAmount > token.balanceOf(address(this))) {
+        if (tradeReturn > token.balanceOf(address(this))) {
```
yudilevi, line 222: balanceOf isn't correct for ETH as it's limited by the amount available for trading.
I suggest a utility function (`amountAvailableForTrading`) that calls balanceOf for tokens, or returns the current ETH sale amount or ETH.

yudilevi (changes requested):
Yup, definitely much better :)

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Fixed by adding the `amountAvailableForTrading` utility function

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: added this check since there is explicit eth sale amount logic embedded in the new function enableTradingETH and also since the new function is used to enable trading for eth, so it doesn't make sense to be able to enable eth trading from both imo

PR #122, opened 2023-11-22
carbon pol eth conversion - price reset and deployment scripts

ivanzhelyazkov:
* Configure the POL upgrade scripts to initialize the eth sale amounts
* Reset the trading start times of ETH at sale amount depletion 
* Set the current sale amount properly if contract balance is below initial sale amount during trade enable

### DIFF
--- a/contracts/pol/CarbonPOL.sol
+++ b/contracts/pol/CarbonPOL.sol
@@ -192,7 +192,7 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
     function enableTradingETH(Price memory price) external onlyAdmin validPrice(price) {
         _tradingStartTimes[NATIVE_TOKEN] = uint32(block.timestamp);
         _initialPrice[NATIVE_TOKEN] = price;
-        _ethSaleAmount.current = uint128(Math.min(address(this).balance, _ethSaleAmount.initial));
+        _ethSaleAmount.current = Math.min(address(this).balance, _ethSaleAmount.initial).toUint128();
         emit TradingEnabled(NATIVE_TOKEN, price);
     }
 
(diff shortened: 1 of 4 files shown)


lbeder (approved):

yudilevi (approved):

lbeder (approved):

contracts/pol/CarbonPOL.sol
```diff
@@ -410,7 +412,9 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
 
         // check if the new sale amount is below the current available eth sale amount
         if (newEthSaleAmount < _ethSaleAmount.current) {
-            _ethSaleAmount.current = newEthSaleAmount;
+            _ethSaleAmount.current = address(this).balance > newEthSaleAmount
```
lbeder, line 415: Super nitpick: you can use Math.min in all these references

contracts/pol/CarbonPOL.sol
```diff
@@ -410,7 +412,9 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
 
         // check if the new sale amount is below the current available eth sale amount
         if (newEthSaleAmount < _ethSaleAmount.current) {
```
lbeder, line 414: Continuing here: but offering more ETH than there is in the contract shouldn't be allowed, no?

ivanzhelyazkov: This is a kind of tricky question since it's possible that the ETH amount in the contract increases - I think limiting the current amount properly is sufficient for our use case.

ivanzhelyazkov: Also this might lead to unexpected behavior since the goal of the function is to set the eth sale amount - not to set it either to contract balance or eth sale amount.

contracts/pol/CarbonPOL.sol
ivanzhelyazkov: Replaced

yudilevi (approved):

lbeder (approved):

PR #123, opened 2023-11-28
carbon pol eth conversion - minor fix

ivanzhelyazkov:
* Flip the input and output amount of the TokenTraded event

### DIFF
--- a/contracts/pol/CarbonPOL.sol
+++ b/contracts/pol/CarbonPOL.sol
@@ -179,7 +179,7 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
         }
         _tradingStartTimes[token] = uint32(block.timestamp);
         _initialPrice[token] = price;
-        emit TradingEnabled(token, price);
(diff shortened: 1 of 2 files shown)


contracts/pol/CarbonPOL.sol
```diff
@@ -308,7 +308,7 @@ contract CarbonPOL is ICarbonPOL, Upgradeable, ReentrancyGuardUpgradeable, Utils
         } else {
             inputAmount = _sellTokenForETH(token, amount);
         }
-        emit TokenTraded(msg.sender, token, amount, inputAmount);
+        emit TokenTraded(msg.sender, token, inputAmount, amount);
```
lbeder, line 311: This is why I always recommend using the key-value notation:

`emit TokenTraded({caller: msg.sender, token: token, inputAmount: inputAmount, ooutputAmount: amount})`

ivanzhelyazkov: Changed to the key-value notation

lbeder (approved):

PR #130, opened 2024-01-22
Extend the supported input range of trade-by-source-amount

barakman:
Reduce the probability of reverting on overflow


yudilevi (approved):

contracts/carbon/Strategies.sol
```diff
@@ -972,6 +977,13 @@ abstract contract Strategies is Initializable {
         }
     }
 
+    function tryAdd(uint256 x, uint256 y) private pure returns (bool safe, uint256 sum) {
```
lbeder: Since you aren't using the `sum` in case `safe` is `false`, you can use OZ's `tryAdd` explicitly: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e86bb45477d335294086c333047626b7fd317e5a/contracts/utils/math/Math.sol#L25

ivanzhelyazkov (approved):

lbeder (changes requested):

contracts/carbon/Strategies.sol
barakman: Yeah, I initially did use that function, but for some reason, it makes the size of the test (helper) contract exceed the maximum, and I wanted to avoid the compilation warning.
The second reason is that we actually have a few other things in there, which we could technically replace with OZ code, so it didn't seem important to follow a line which we aren't enforcing anyway.
P.S.: if OZ had given it just a little bit more thought, then they would have probably ended up doing it the exact same way we did it here.

lbeder: The size of test contracts isn't really an issue. They will eventually become too large, even without the fix. DIY or code duplicate is always undesired. If there are other places like this, IMHO they should be refactored just as well. Less DIY/duplication == easier to review and reason about the code.

@yudilevi your call

barakman: We also don't use OZ's `mulDiv` (which they've apparently added at some point), and we can replace our `mulDivF` and `mulDivC` with (all across the code of course, otherwise, we get two different versions of each one of these functions).

yudilevi: I’d probably go with OZ’s code for any new code (no muldiv replacement yet)

barakman: OK, please note that OZ's implementation is slightly more expensive than ours:
- OZ implementation: 131 gas in case of an overflow and 117 gas in case of no overflow
- Our implementation: 38 gas regardless of whether an overflow occurs or doesn't occur

This means:
- 79 gas **per order** per trade in most cases (those in which an overflow doesn't occur)
- 93 gas **per order** per trade in rare cases (those in which an overflow is handled)

Not significant but still something to consider...

lbeder: IMHO, insignificant and can be ignored.

barakman: Done

yudilevi (approved):

contracts/carbon/Strategies.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity 0.8.19;
 import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 import { EnumerableSetUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/structs/EnumerableSetUpgradeable.sol";
 import { MathUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/math/MathUpgradeable.sol";
+import { SafeMathUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/math/SafeMathUpgradeable.sol";
```
lbeder: Not important, but since these libraries aren't really "upgradeable" (they are equivalent to the non-upgradeable versions), you may consider importing them as `SafeMath` in the future (i.e., `import { SafeMathUpgradeable as SafeMath } ...`

contracts/carbon/Strategies.sol
```diff
@@ -779,7 +781,7 @@ abstract contract Strategies is Initializable {
         uint256 temp4 = MathEx.mulDivC(temp1, temp1, factor);
         uint256 temp5 = MathEx.mulDivC(temp3, A, factor);
 
-        (bool safe, uint256 sum) = tryAdd(temp4, temp5);
+        (bool safe, uint256 sum) = SafeMathUpgradeable.tryAdd(temp4, temp5);
```
lbeder: Since you've declared `using SafeMathUpgradeable for uint256` above, you should do `temp4.tryAdd(temp5)`. BTW, this is probably why you've experienced a small increase in contract size, but it's ok, since this is how most of other libraries (e.g., `SafeCastUpgradeable`) are used. 

BTW, `MathUpgradeable` should be used this way as well

lbeder (changes requested):

contracts/carbon/Strategies.sol
barakman: IMO, `temp4.tryAdd(temp5)` is less clear than `Math.tryAdd(temp4, temp5)`.

Same goes for using `MathEx.mulDivC(x, y, z)` instead of `x.mulDivC(y, `)`, etc.

Replaced all 3 "upgradable" libraries with their "non-upgradable" equivalence:
```
import { Math } from "@openzeppelin/contracts/utils/math/Math.sol";
import { SafeMath } from "@openzeppelin/contracts/utils/math/SafeMath.sol";
import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
```

lbeder: `temp4.tryAdd(temp5)` is how similar libraries are used in other places, so this is the way to be consistent

barakman: For the same reason, I've used `Math.max(factor1, factor2)` and not `factor1.max(factor2)`.

Same goes for using `MathEx.mulDivC(x, y, z)` instead of `x.mulDivC(y, `)`, etc.

So if anything, using `temp4.tryAdd(temp5)` will be inconsistent with respect to the rest of the code.

lbeder: Remember using `SafeMath` for `add`, `sub`, `mul`, etc.? It was always `x.add(y)` and not `add(x, y)`. In any case, let's let @yudilevi to decide on this one

cache_forge/solidity-files-cache.json
```diff
@@ -0,0 +1,7048 @@
+{
```
lbeder: @barakman note that this shouldn't be committed too

contracts/carbon/Strategies.sol
yudilevi: If we're using `using`, we should use the standard `using` method.
If we don't want to use the standard method, we shouldn't be using `using`.
So let's pick one method and stick to the standard.

lbeder: @barakman make your choice and I'll quickly review and approve ❤️

barakman: I have previously removed a couple of `using` statements which were not used in the contract:
- `using Math`
- `using SafeMath`

While leaving in place the `using` statement which WAS used in the contract:
- `using SafeCast`

If that's what you meant, then it's already done.

lbeder: Yeah, this is the first approach

yudilevi (approved):

lbeder (approved):

PR #132, opened 2024-01-31
setup multiple network deployment

ivanzhelyazkov:
* Setup multiple network deployments for Mainnet, Base, Arbitrum and Mantle
* Setup tenderly forking for the above networks
* Setup tenderly testnet creation and account funding for Mainnet and Base
* Modify Upgradeable contract to be able to set the version on new deployments to the correct version (in line with the original version)


yudilevi (approved):
1 failed test, otherwise looks great

ivanzhelyazkov:
> 1 failed test, otherwise looks great

Fixed

data/named-accounts.ts
```diff
@@ -1,73 +1,182 @@
...
-    [DeploymentNetwork.TenderlyTestnet]: address
-});
+interface EnvOptions {
+    TENDERLY_NETWORK_NAME?: string;
+}
 
-const rinkeby = (address: string) => ({
-    [DeploymentNetwork.Rinkeby]: address
-});
+const { TENDERLY_NETWORK_NAME = 'mainnet' }: EnvOptions = process.env as any as EnvOptions;
```
lbeder: Should be `DeploymentNetwork.Mainnet`, no?

data/named-accounts.ts
```diff
@@ -1,73 +1,182 @@
...
+
+const TENDERLY_NETWORK_ID = chainIds[TENDERLY_NETWORK_NAME as keyof typeof chainIds];
+
+const mainnet = (address: string) => {
+    if (TENDERLY_NETWORK_ID === chainIds['mainnet']) {
```
lbeder: Same here (and below)

data/named-accounts.ts
```diff
@@ -1,73 +1,182 @@
...
+            [DeploymentNetwork.TenderlyTestnet]: address
+        };
+    }
+    return {
+        [DeploymentNetwork.Mainnet]: address
+    };
+};
+
+const base = (address: string) => {
+    if (TENDERLY_NETWORK_ID === chainIds['base']) {
```
lbeder: `DeploymentNetwork.Base`

data/named-accounts.ts
```diff
@@ -1,73 +1,182 @@
...
+    }
+    return {
+        [DeploymentNetwork.Mantle]: address
+    };
+};
 
 const TestNamedAccounts = {
     ethWhale: {
-        ...mainnet('0xDA9dfA130Df4dE4673b89022EE50ff26f6EA73Cf')
+        ...getAddress(mainnet, '0xDA9dfA130Df4dE4673b89022EE50ff26f6EA73Cf'),
```
lbeder, line 79: I'm not sure why the new syntax is needed (seems unnecessarily more verbose than what is was before). A workaround to some TypeScript issue?

data/named-accounts.ts
```diff
@@ -1,73 +1,182 @@
...
+    foundationMultisig: { ...getAddress(mainnet, '0xeBeD45Ca22fcF70AdCcAb7618C51A3Dbb06C8d83') },
+    foundationMultisig2: { ...getAddress(mainnet, '0x0c333d48Af19c2b42577f3C8f4779F0347F8C819') },
+    daoMultisig: { ...getAddress(mainnet, '0x7e3692a6d8c34a762079fa9057aed87be7e67cb8') },
+    daoPauserMultisig: { ...getAddress(mainnet, '0xc140c1CD2e587fC04DAa780d49b616F768476660') },
 
     ...TokenNamedAccounts,
     ...TestNamedAccounts,
     ...BancorNamedAccounts
-};
+};
```
lbeder: Missing new line at the end of the while. You should integrate/run the formatter.

deploy/scripts/base/0002-Voucher.ts
```diff
@@ -0,0 +1,21 @@
...
+    const { deployer } = await getNamedAccounts();
+
+    await deployProxy(
+        {
+            name: InstanceName.Voucher,
+            from: deployer
+        },
+        {
+            args: [true, 'ipfs://QmUyDUzQtwAhMB1hrYaQAqmRTbgt9sUnwq11GeqyzzSuqn', '']
+        }
```
lbeder, line 16: Same URI for all chains, right?

utils/Constants.ts
```diff
@@ -8,18 +8,30 @@ const {
...
+    Base = 'base',
+    Arbitrum = 'arbitrum',
+    Sepolia = 'sepolia',
     Hardhat = 'hardhat',
     Tenderly = 'tenderly',
     TenderlyTestnet = 'tenderly-testnet'
 }
 
+export enum NetworkId {
+    Mainnet = '1',
```
lbeder, line 21: Chain ids should be `bigint` (this is how they encoded/decoded/requested by various APIs), but it's not the of the world and can be kept as a string in this PR

utils/Deploy.ts
```diff
@@ -221,7 +232,7 @@ const logParams = async (params: FunctionParams) => {
 
     for (const [i, arg] of args.entries()) {
         const input = fragment.inputs[i];
-        Logger.log(`    ${input.name} (${input.type}): ${arg.toString()}`);
+        Logger.log(`    ${input?.name} (${input?.type}): ${arg?.toString()}`);
```
lbeder: An input shouldn't be undefined here, but in any case, just check that it is and `continue` so that you don't need to `input?` everywhere

lbeder (changes requested):

data/named-accounts.ts
ivanzhelyazkov: Yes, typescript issue related with the type hardhat config's "namedAccounts" field expects. Fails to install packages otherwise.

data/named-accounts.ts
ivanzhelyazkov: fixed

data/named-accounts.ts
ivanzhelyazkov: fixed

data/named-accounts.ts
ivanzhelyazkov: fixed

data/named-accounts.ts
ivanzhelyazkov: fixed

deploy/scripts/base/0002-Voucher.ts
ivanzhelyazkov: added a constant

utils/Constants.ts
ivanzhelyazkov: going to keep them as strings for now can fix at a later stage

utils/Deploy.ts
ivanzhelyazkov: fixed

lbeder (approved):

PR #133, opened 2024-02-02
Voucher immutability

ivanzhelyazkov:
* Replace voucher minter role with controller address, settable only once
* Add tests
* Add upgrade scripts


contracts/voucher/Voucher.sol
```diff
@@ -35,8 +32,14 @@ contract Voucher is IVoucher, Upgradeable, ERC721Upgradeable, Utils {
     // a mapping between an owner to its tokenIds
     mapping(address => EnumerableSet.UintSet) internal _ownedTokens;
 
+    // controller address - used to mint / burn
+    address private _controller;
+
+    // a flag indicating whether the controller role has been set
+    bool private _controllerSet;
```
lbeder: Why not just checking that `_controller` isn't `0x0`?

contracts/voucher/Voucher.sol
```diff
@@ -233,6 +226,33 @@ contract Voucher is IVoucher, Upgradeable, ERC721Upgradeable, Utils {
...
+    /**
+     * @dev sets the controller address
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of this contract
+     * - controller address must not be set
+     */
+    function setController(address controllerAddress) public onlyAdmin {
+        if (_controllerSet) {
```
lbeder: A few notes:
- You don't check that `controllerAddress` isn't `0x0` (+ see the comment above about not needing `_controllerSet`)
- `setController` seems a bit misplaced. I'd expect, in a contract like this:
    - For new deployments, the controller is passed a parameter to `initialize` (and this way can be only set once as well).
    - For the current instance, the controller is set via `postUpgrade` (passed externally). If you had kept the `ROLE_MINTER` role, then `postUpgrade` could've had enumerated it and assigned the controller to be the first address with this rule (which is possible since it uses `EnumerableSet`).
- The function can be `external`

contracts/voucher/interfaces/IVoucher.sol
```diff
@@ -9,12 +9,15 @@ import { IUpgradeable } from "../../utility/interfaces/IUpgradeable.sol";
  * @dev Voucher interface
  */
 interface IVoucher is IUpgradeable, IERC721Upgradeable {
```
lbeder, line 11: It may be useful to expose a `controller()` public getter (but it's up to you)

lbeder (changes requested):

contracts/voucher/interfaces/IVoucher.sol
ivanzhelyazkov: added

contracts/voucher/Voucher.sol
ivanzhelyazkov: changed

contracts/voucher/Voucher.sol
ivanzhelyazkov: changed 1 and 3, for 2) it's not necessary, since the upgradeability / proxies are going to be removed

lbeder (approved):

yudilevi (approved):

PR #134, opened 2024-02-16
CarbonVortex - add withdraw funds

ivanzhelyazkov:
* Add admin-controlled withdrawFunds function to CarbonVortex
* Add tests
* Add upgrade scripts

### DIFF
--- a/contracts/vortex/CarbonVortex.sol
+++ b/contracts/vortex/CarbonVortex.sol
@@ -101,7 +101,7 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
      * @inheritdoc Upgradeable
      */
     function version() public pure override(IVersioned, Upgradeable) returns (uint16) {
-        return 2;
+        return 3;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


contracts/vortex/CarbonVortex.sol
```diff
@@ -183,6 +183,28 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
+     */
+    function withdrawFunds(
+        Token token,
+        address payable target,
+        uint256 amount
+    ) external validAddress(target) nonReentrant onlyAdmin {
```
barakman, line 197: `onlyAdmin` by itself takes care of reentrancy, making `nonReentrant` somewhat redundant (in terms of gas).

yudilevi: It’s a good practice, gas is irrelevant here

lbeder: `onlyAdmin` doesn't take care of reentrancy in general. Imagine that the contract sends ETH to the owner `msg.sender` who is incidentally a contract that reenters the function...

lbeder: But there is no real reentrancy concern here anyway, but as @yudilevi said - better safe then sorry

contracts/vortex/CarbonVortex.sol
```diff
@@ -183,6 +183,28 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
     }
 
+    /**
+     * @dev withdraws funds held by the contract and sends them to an account
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
+     */
+    function withdrawFunds(
```
lbeder, line 193: I suggest adding a batch function, to save gas when this is being invoked by an external sweeper contract

yudilevi: discussed and decided that for the time being it's unnecessary and that it will be added to a generic Vault contract later on

yudilevi (approved):

lbeder (approved):

barakman (approved):

PR #138, opened 2024-03-05
Remove pause

yudilevi:


deploy/scripts/mainnet/0013-CarbonController-upgrade.ts
```diff
@@ -0,0 +1,22 @@
+import { DeployedContracts, InstanceName, setDeploymentMetadata, upgradeProxy } from '../../../utils/Deploy';
+import { DeployFunction } from 'hardhat-deploy/types';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+
+/**
+ * @dev trade by source extended input upgrade
```
ivanzhelyazkov: comment is incorrect - should be remove pausability upgrade

ivanzhelyazkov (approved):
Looks good.
Could consider removing `PausableUpgradeable` from the inheritance chain (might be too much of a hassle due to possibly opening up storage collisions)

contracts/carbon/CarbonController.sol
```diff
@@ -26,9 +26,6 @@ contract CarbonController is
     OnlyProxyDelegate,
     Utils
 {
-    // the emergency manager role is required to pause/unpause
-    bytes32 private constant ROLE_EMERGENCY_STOPPER = keccak256("ROLE_EMERGENCY_STOPPER");
```
lbeder, line 30: You should either consider explicitly revoking this role from existing accounts or documenting this role as deprecating (i.e. so that it's clear that this remains set for existing contracts and should never be reused, etc.).

contracts/carbon/CarbonController.sol
```diff
@@ -26,9 +26,6 @@ contract CarbonController is
     OnlyProxyDelegate,
     Utils
 {
-    // the emergency manager role is required to pause/unpause
```
lbeder, line 29: `PausableUpgradeable` is no longer needed, but removing it will indeed affect the storage layout. One way to resolve it is to add a `bool private reserved0` (+ gap) on top of the storage (either explicitly or by replacing `PausableUpgradeable` with `ReservedXXX` that only reserved `50` slots and that's it

hardhat.config.ts
```diff
@@ -35,7 +35,7 @@ interface EnvOptions {
 const {
     TENDERLY_TESTNET_PROVIDER_URL = '',
     VERIFY_API_KEY = '',
-    GAS_PRICE: gasPrice = 'auto',
+    GAS_PRICE = 'auto',
```
lbeder: Unrelated to this PR, but you can also consider using v2 transactions by setting max base and priority fees instead (== it'll cheapen the deployments, since currently both are effectively set to the same value)

lbeder: Why did you remove the alias of `gasPrice`? Without it, you indeed have to duplicate `gasPrice: GAS_PRICE` multiple times below...

lbeder (changes requested):

deploy/scripts/mainnet/0013-CarbonController-upgrade.ts
yudilevi: This will require a storage slot change, so we'd rather keep it.
We could replace it with another contract that has similar storage layout but it's an unnecessary risk.

contracts/carbon/CarbonController.sol
yudilevi: it will be revoked from existing accounts.

contracts/carbon/CarbonController.sol
yudilevi: thought it's a bit risky, but it's definitely doable.

hardhat.config.ts
yudilevi: @ivanzhelyazkov that's to you

ivanzhelyazkov: fixed

contracts/carbon/CarbonController.sol
lbeder: As part of the upgrade script?

contracts/carbon/CarbonController.sol
lbeder: It is riskier than keeping it as is, but:
1. It's easy to verify that the layout was preserved (do we have the layout export command here as well?)
2. It cleans the code a bit and doesn't keep unused/unneeded methods (but it's not the end of the world)

contracts/carbon/CarbonController.sol
yudilevi: nope, before the upgrade

contracts/carbon/CarbonController.sol
yudilevi: removed (tested with the export command)

lbeder (approved):

deploy/scripts/mainnet/0013-CarbonController-upgrade.ts
yudilevi: fixed

contracts/carbon/CarbonController.sol
```diff
@@ -22,18 +21,18 @@ contract CarbonController is
...
-    bytes32 private constant ROLE_EMERGENCY_STOPPER = keccak256("ROLE_EMERGENCY_STOPPER");
-
     // the fees manager role is required to withdraw fees
     bytes32 private constant ROLE_FEES_MANAGER = keccak256("ROLE_FEES_MANAGER");
 
     uint16 private constant CONTROLLER_TYPE = 1;
 
+    // deprecated parent storage vars
+    bool private _deprecated;
+    uint256[49] private __deprecated;
```
barakman, line 34: Why not just `uint256[MAX_GAP] private __deprecated;`?
Also, for readability, probably better placed after all constant and immutable variables.

yudilevi: Because I wanted to keep the exact definition from the removed contract

ivanzhelyazkov (approved):

PR #139, opened 2024-04-03
Increase the maximum permitted value of `timeElapsed / halfLife` from ~23 to 128

barakman:
The calculation of `2 ^ x` for a non-integer value of `x` is limited by `x < 16 * ln(2)`, which is approximately 23.08.

The solution to this problem relies on the identity:
```
A ^ B.C = A ^ (B + 0.C) = A ^ B * A ^ 0.C
```
For example:
```
2 ^ 3.4 = 2 ^ (3 + 0.4) = 2 ^ 3 * 2 ^ 0.4
```
Calculating `2 ^ 3` is easy, and calculating `2 ^ 0.4` is also easy, because `0.4` is way below the limit (23.08).

### DIFF
--- a/contracts/utility/ExpDecayMath.sol
+++ b/contracts/utility/ExpDecayMath.sol
@@ -12,7 +12,7 @@ library ExpDecayMath {
      * @dev returns the amount required for a token after a given time period since trading has been enabled
      *
      * the returned value is calculated as `amount / 2 ^ (timeElapsed / halfLife)`
-     * note that the input value to this function is limited by `timeElapsed / halfLife <= 128`
+     * note that the input value to this function is limited by `timeElapsed / halfLife < 129`
      */
     function calcExpDecay(uint256 amount, uint32 timeElapsed, uint32 halfLife) internal pure returns (uint256) {
         uint256 integerPart = timeElapsed / halfLife;
--- a/contracts/utility/ExpDecayMath.sol
+++ b/contracts/utility/ExpDecayMath.sol
@@ -12,12 +12,13 @@ library ExpDecayMath {
      * @dev returns the amount required for a token after a given time period since trading has been enabled
      *
      * the returned value is calculated as `amount / 2 ^ (timeElapsed / halfLife)`
-     * note that because the exponentiation function is limited to an input of up to (and excluding)
-     * 16 / ln 2, the input value to this function is limited by `timeElapsed / halfLife < 16 / ln 2`
+     * note that the input value to this function is limited by `timeElapsed / halfLife <= 128`
      */
     function calcExpDecay(uint256 amount, uint32 timeElapsed, uint32 halfLife) internal pure returns (uint256) {
-        Fraction memory input = Fraction({ n: timeElapsed, d: halfLife });
+        uint256 integerPart = timeElapsed / halfLife;
+        uint256 fractionPart = timeElapsed % halfLife;
+        Fraction memory input = Fraction({ n: fractionPart , d: halfLife });
         Fraction memory output = MathEx.exp2(input);
-        return MathEx.mulDivF(amount, output.d, output.n);
+        return MathEx.mulDivF(amount, output.d, output.n * 2 ** integerPart);
     }
 }
--- a/test/math/ExpDecayMath.ts
+++ b/test/math/ExpDecayMath.ts
@@ -12,7 +12,7 @@ describe('ExpDecayMath', () => {
     let expDecayMath: TestExpDecayMath;
     const ONE = new Decimal(1);
     const TWO = new Decimal(2);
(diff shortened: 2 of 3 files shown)


ivanzhelyazkov (approved):

test/math/ExpDecayMath.ts
```diff
@@ -33,7 +33,7 @@ describe('ExpDecayMath', () => {
                 });
             } else {
                 await expect(expDecayMath.calcExpDecay(ethAmount, timeElapsed, halfLife)).to.revertedWithError(
-                    'Overflow'
+                    'panic code 0x11'
```
lbeder, line 36: Should be `revertedWithPanic`

contracts/utility/ExpDecayMath.sol
```diff
@@ -12,12 +12,13 @@ library ExpDecayMath {
...
+     * note that the input value to this function is limited by `timeElapsed / halfLife <= 128`
      */
     function calcExpDecay(uint256 amount, uint32 timeElapsed, uint32 halfLife) internal pure returns (uint256) {
-        Fraction memory input = Fraction({ n: timeElapsed, d: halfLife });
+        uint256 integerPart = timeElapsed / halfLife;
+        uint256 fractionPart = timeElapsed % halfLife;
+        Fraction memory input = Fraction({ n: fractionPart , d: halfLife });
         Fraction memory output = MathEx.exp2(input);
-        return MathEx.mulDivF(amount, output.d, output.n);
+        return MathEx.mulDivF(amount, output.d, output.n * 2 ** integerPart);
```
lbeder, line 22: Are you sure this won't revert in edge case? I.e., if `output.n * 2 ** integerPart` become too huge?

barakman: I want it to.
Otherwise I would have used `output.n << integerPart`.
BTW, regardless of this change, `mulDivF` itself reverts whenever its final outcome is (still) larger than 256 bits.
The purpose in this PR is to extend the supported input range, not to make it equal to the entire input range.

test/math/ExpDecayMath.ts
barakman: I'll give it a try, TX!

barakman: `Error: Invalid Chai property: revertedWithPanic`

lbeder: Check the documentation: https://hardhat.org/hardhat-chai-matchers/docs/overview: you need to use @nomicfoundation/hardhat-chai-matchers instead of @nomiclabs/hardhat-waffle.

If this is too complicated, then leave it as is.

contracts/utility/ExpDecayMath.sol
```diff
@@ -12,12 +12,13 @@ library ExpDecayMath {
      * @dev returns the amount required for a token after a given time period since trading has been enabled
      *
      * the returned value is calculated as `amount / 2 ^ (timeElapsed / halfLife)`
-     * note that because the exponentiation function is limited to an input of up to (and excluding)
-     * 16 / ln 2, the input value to this function is limited by `timeElapsed / halfLife < 16 / ln 2`
+     * note that the input value to this function is limited by `timeElapsed / halfLife <= 128`
      */
     function calcExpDecay(uint256 amount, uint32 timeElapsed, uint32 halfLife) internal pure returns (uint256) {
-        Fraction memory input = Fraction({ n: timeElapsed, d: halfLife });
+        uint256 integerPart = timeElapsed / halfLife;
```
ivanzhelyazkov, line 18: Thoughts on adding:
```
        if (integerPart > 128)
            revert ExpDecayOverflow();
```

might be better to trace where the error appears

barakman: The more accurate condition to assert, and even directly on the input, is:
```
timeElapsed < halfLife * 129
```
But note that this restriction is tightly-coupled with the implementation of the `exp2` function.
I therefore did not enforce it in the `expDecay` function, just like we don't normally impose these checks on function input in most cases, leaving it for the EVM to handle during runtime.
Let me know if you prefer otherwise (will require adapting the expected error-message in the tests).

Also note that `halfLife * 129` by itself can overflow, so the exact condition is technically:
```
timeElapsed < uint256(halfLife) * 129
```
Of course, when `halfLife * 129` overflows, the expression `timeElapsed < uint256(halfLife) * 129` is false, so it is only a matter of where you prefer to receive the error from (an overflow or a standard `revert` statement).

ivanzhelyazkov: on second thought, might not be necessary - might be better to leave it as is

ivanzhelyazkov (approved):

PR #140, opened 2024-04-10
Carbon Vortex 2.0

ivanzhelyazkov:
* Add Carbon Vortex 2.0 contract
* Add Carbon Vortex 2.0 tests

Based on this governance proposal:
https://gov.bancor.network/t/carbon-vortex-2-0-dutch-auction/4597


contracts/utility/ExpDecayMath.sol
```diff
@@ -12,12 +12,13 @@ library ExpDecayMath {
      * @dev returns the amount required for a token after a given time period since trading has been enabled
      *
      * the returned value is calculated as `amount / 2 ^ (timeElapsed / halfLife)`
-     * note that because the exponentiation function is limited to an input of up to (and excluding)
-     * 16 / ln 2, the input value to this function is limited by `timeElapsed / halfLife < 16 / ln 2`
+     * note that the input value to this function is limited by `timeElapsed / halfLife <= 128`
```
barakman: Not aligned with latest changes in the ExpDecay PR

contracts/vortex/interfaces/ICarbonVortex.sol
```diff
@@ -11,50 +11,186 @@ interface ICarbonVortex is IUpgradeable {
...
+    /**
+     * @notice triggered after the price updates for a token
+     */
+    event PriceUpdated(Token indexed token, Price price);
 
     /**
      * @dev triggered when tokens have been withdrawn by the admin
      */
-    event FundsWithdrawn(Token indexed token, address indexed caller, address indexed target, uint256 amount);
+    event FundsWithdrawn(Token[] tokens, uint256[] amounts, address indexed caller, address indexed target);
```
barakman: Maybe place the indexed parameters first, in order to preserve our general convention?

contracts/vortex/interfaces/ICarbonVortex.sol
```diff
@@ -11,50 +11,186 @@ interface ICarbonVortex is IUpgradeable {
...
+    /**
+     * @notice returns the source amount required given a target amount
+     */
+    function expectedTradeInput(Token token, uint128 targetAmount) external view returns (uint128 sourceAmount);
+
+    /**
+     * @notice returns the current token price (targetToken / TKN)
+     * @notice if token == targetToken, returns finalTargetToken / targetToken price
+     */
+    function tokenPrice(Token token) external view returns (Price memory price);
```
barakman: Minor: output argument name declared in the last 3 functions (in opposed to all the rest).

contracts/vortex/CarbonVortex.sol
```diff
@@ -3,58 +3,102 @@ pragma solidity 0.8.19;
...
+    // first token for swapping
+    Token private immutable _targetToken;
+    // second (optional) token for swapping
+    Token private immutable _finalTargetToken;
+
+    // total target (if no finalTarget token is defined) / finalTarget tokens collected in transferAddress
+    uint256 private _totalCollected;
 
     // rewards percentage
-    uint256 private _rewardsPPM;
```
barakman, line 42: Shouldn't we keep in the contract something like:
```
uint256 private _totalBurned_deprecated;
uint256 private _rewardsPPM_deprecated;
```
Then add all the new state variables and update the size of `__gap` accordingly?

contracts/vortex/CarbonVortex.sol
```diff
@@ -105,31 +172,171 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
+     */
+    function withdrawFunds(
+        Token[] calldata tokens,
+        address payable target,
+        uint256[] calldata amounts
+    ) external validAddress(target) validateTokens(tokens) nonReentrant onlyAdmin {
```
barakman, line 298: The `onlyAdmin` modifier already takes care of reentrancy so the `nonReentrant` modifier is redundant here.

But if you insist on having both of them, then there's a bunch of functions above guarded only by the first one.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+
             // get reward amount for token
-            rewardAmounts[i] = MathEx.mulDivF(balances[i], rewardsPercentage, PPM_RESOLUTION);
+            rewardAmounts[i] = MathEx.mulDivF(totalFeeAmount, rewardsPercentage, PPM_RESOLUTION);
         }
 
-        // convert tokens to BNT
-        for (uint256 i = 0; i < len; i = uncheckedInc(i)) {
+        // go through all tokens and start / reset dutch auction if necessary
+        for (uint256 i = 0; i < tokens.length; ++i) {
```
barakman: Can use `uncheckedInc(i)` here. And in fact, also `len`, so unless I'm missing anything, you could just keep the previous version of that loop-header.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
                 continue;
             }
-
-            // get trade amount for token
-            uint256 tradeAmount;
-            unchecked {
-                // safe because balances[i] >= rewardAmounts[i]
-                tradeAmount = balances[i] - rewardAmounts[i];
+            // skip the final target token
+            if (token == _finalTargetToken) {
```
barakman, line 393: Probably best to do that first, since `_disabledPairs[token]` yields a storage-read.
In fact, you can just add this check into the previous check, right before that storage-read.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+                    _amountAvailableForTrading(token) < _minTokenSaleAmounts[token] ||
+                    _auctionPriceIsBelowMinimum(token)
+                ) {
+                    // reset trading for target token
+                    _resetTradingTarget(rewardAmount);
+                }
+            } else {
+                if (
+                    !_tradingEnabled(token) ||
+                    _amountAvailableForTrading(token) - feeAmount < _minTokenSaleAmounts[token] ||
```
barakman: If it is not guaranteed that `_amountAvailableForTrading(token) >= feeAmount`, then it is probably better to do:
```
_amountAvailableForTrading(token) < _minTokenSaleAmounts[token] + feeAmount
```
Since it pretty much guarantees that this check will never revert the transaction on arithmetic overflow.

contracts/vortex/CarbonVortex.sol
```diff
@@ -252,37 +858,109 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
 
-    function _validateTokens(Token[] calldata tokens) private view {
+    /**
+     * @dev returns the token amount available for trading
+     */
+    function _amountAvailableForTrading(Token token) private view returns (uint128) {
+        if (token == _targetToken) {
+            return _targetTokenSaleAmount.current;
+        } else {
+            return uint128(token.balanceOf(address(this)));
```
barakman: Unsafe casting.
You should probably use `SafeCast.toUint128()` instead.

contracts/vortex/CarbonVortex.sol
```diff
@@ -252,37 +858,109 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+        // get time elapsed since trading was enabled
+        uint32 timeElapsed = uint32(block.timestamp) - tradingStartTime;
+        // get the halflife for the token
+        uint32 currentPriceDecayHalfLife = token == _targetToken ? _targetTokenPriceDecayHalfLife : _priceDecayHalfLife;
+        // check if the maximum amount of halflifes have been reached
+        if (timeElapsed / currentPriceDecayHalfLife >= 128) {
+            return true;
+        } else {
+            return false;
         }
```
barakman: Should be:
```
// check if the time-elapsed has exceeded the maximum time period supported for the given half-life
return (uint256(timeElapsed) >= uint256(currentPriceDecayHalfLife) * 129;
```
The rewards-plan can continue running without an overflow for as long as `timeElapsed < HalfLife * 129`.
The casting to `uint256` is in order to avoid an overflow (of 32 bits) in the check itself.

contracts/vortex/CarbonVortex.sol
```diff
@@ -252,37 +858,109 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+     */
+    function _allocateRewards(address sender, Token[] memory tokens, uint256[] memory rewardAmounts) private {
+        uint256 len = tokens.length;
+        for (uint256 i = 0; i < len; i = uncheckedInc(i)) {
+            Token token = tokens[i];
+            uint256 rewardAmount = rewardAmounts[i];
+            // transfer the rewards to caller
+            // safe due to nonReentrant modifier (forwards all available gas in case of ETH)
+            token.unsafeTransfer(sender, rewardAmount);
         }
```
barakman, line 960: This function is only called from function `execute`, so it might be worth implementing it directly inside that function. Perhaps even as part of the loop over tokens in there (provided that we can indeed allocate rewards on the fly and not only at the end of that loop).
It would also make it easier to understand the comment about being safe due to `nonReentrant` modifier.

test/math/ExpDecayMath.ts
```diff
@@ -13,6 +12,7 @@ describe('ExpDecayMath', () => {
     let expDecayMath: TestExpDecayMath;
     const ONE = new Decimal(1);
     const TWO = new Decimal(2);
+    const MAX = new Decimal(128);
```
barakman: Not aligned with latest changes in the ExpDecay PR

test/math/ExpDecayMath.ts
```diff
@@ -21,7 +21,7 @@ describe('ExpDecayMath', () => {
     const calcExpDecay = (ethAmount: BigNumberish, timeElapsed: number, halfLife: number) => {
         it(`calcExpDecay(${ethAmount}, ${timeElapsed}, ${halfLife})`, async () => {
             const f = new Decimal(timeElapsed).div(halfLife);
-            if (f.lt(EXP2_INPUT_TOO_HIGH)) {
+            if (f.lte(MAX)) {
```
barakman: Not aligned with latest changes in the ExpDecay PR

test/math/ExpDecayMath.ts
```diff
@@ -33,7 +33,7 @@ describe('ExpDecayMath', () => {
                 });
             } else {
                 await expect(expDecayMath.calcExpDecay(ethAmount, timeElapsed, halfLife)).to.revertedWithError(
-                    'Overflow'
+                    'panic code 0x11'
                 );
             }
         });
```
barakman, line 39: A few more tests missing below (not aligned with latest changes in the ExpDecay PR)

test/forge/VortexTestCaseParser.t.sol
```diff
@@ -0,0 +1,111 @@
...
+
+    /// @dev convert a price at timestamp struct to uint256
+    function convertPriceAtTimestampToUint(
+        PriceAtTimestampString memory priceAtTimestampString
+    ) private pure returns (PriceAtTimestamp memory priceAtTimestamp) {
+        return
+            PriceAtTimestamp({
+                timestamp: uint32(stringToUint(priceAtTimestampString.timestamp)),
+                sourceAmount: uint128(stringToUint(priceAtTimestampString.sourceAmount)),
+                targetAmount: uint128(stringToUint(priceAtTimestampString.targetAmount))
```
barakman: Unsafe casts.
Should probably use `SafeCast.toUint32()` and `SafeCast.toUint128()`.

contracts/vortex/CarbonVortex.sol
```diff
@@ -105,31 +172,171 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
 
     /**
-     * @inheritdoc ICarbonVortex
+     * @notice sets the rewards ppm
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
      */
-    function setRewardsPPM(uint256 newRewardsPPM) external onlyAdmin validFee(uint32(newRewardsPPM)) {
```
barakman, line 110: Wow, previous version is really unsafe, because of that unsafe cast in there.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+        _initialPrice[_targetToken] = price;
+        _targetTokenSaleAmount.current = Math
+            .min(_targetToken.balanceOf(address(this)) - rewardAmount, _targetTokenSaleAmount.initial)
+            .toUint128();
+        // set price decay halflife to the current price decay halflife
+        _setTargetTokenPriceDecayHalfLife(_priceDecayHalfLife);
+        // emit trading reset event
+        emit TradingReset({ token: _targetToken, price: price });
+    }
+
```
barakman, line 463: Functions `_resetTrading` and `_resetTradingTarget` share identical code.
Consider reusing (i.e., in a separate function, and only if you find it a good suggestion).

barakman (changes requested):
FYI: `uint32(block.timestamp)` is an unsafe cast which appears in about 5 different places in the code, but I suppose (know) that it is harmless in the foreseeable future...

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+     */
+    function _resetTrading(Token token, uint256 rewardAmount) private {
+        Price memory price = Price({
+            sourceAmount: INITIAL_PRICE_SOURCE_AMOUNT,
+            targetAmount: INITIAL_PRICE_TARGET_AMOUNT
+        });
+        _tradingStartTimes[token] = uint32(block.timestamp);
+        _initialPrice[token] = price;
+        // set min token sale amount
+        _setMinTokenSaleAmount(token, (token.balanceOf(address(this)) - rewardAmount).toUint128() / 2);
```
barakman, line 444: Pretty thin, but dividing by 2 before the cast to `uint128` instead of after, could be the difference between a successful transaction and one which reverts on overflow.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+        } else {
+            // if target token is not native, transfer the tokens to the contract
+            _targetToken.safeTransferFrom(msg.sender, address(this), sourceAmount);
+        }
+        // transfer the tokens to caller
+        token.safeTransfer(msg.sender, targetAmount);
+
+        // if the target token is native, refund any excess native token to caller
+        if (_targetToken == NATIVE_TOKEN && msg.value > sourceAmount) {
+            payable(msg.sender).sendValue(msg.value - sourceAmount);
```
barakman, line 630: Minor: the computation of `msg.value - sourceAmount` can be unchecked.
Also, consider moving this refund part a few lines above, where you already know that this is the native token (in the `else` part, right after calling `safeTransferFrom`).

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+            revert InvalidTrade();
+        }
+
+        // check enough final target token (if final target token is native) has been sent for the trade
+        if (_finalTargetToken == NATIVE_TOKEN) {
+            if (msg.value < sourceAmount) {
+                revert InsufficientNativeTokenSent();
+            } else {
+                // transfer the tokens to the _transferAddress
+                payable(_transferAddress).sendValue(sourceAmount);
```
barakman: What about returning any leftover to the caller, i.e., if `msg.value > sourceAmount`?

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+            _finalTargetToken.safeTransferFrom(msg.sender, _transferAddress, sourceAmount);
+        }
+
+        // transfer the _targetToken to the user
+        // safe due to nonReentrant modifier (forwards all available gas if native)
+        _targetToken.unsafeTransfer(msg.sender, targetAmount);
+
+        // if final target token is native, refund any excess native token to caller
+        if (_finalTargetToken == NATIVE_TOKEN && msg.value > sourceAmount) {
+            payable(msg.sender).sendValue(msg.value - sourceAmount);
```
barakman, line 673: Minor: the computation of `msg.value - sourceAmount` can be unchecked.
Also, consider moving this refund part a few lines above, where you already know that this is the native token (after you transfer the tokens to the `_transferAddress`).

barakman (changes requested):

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+                    _auctionPriceIsBelowMinimum(token)
+                ) {
+                    // reset trading for target token
+                    _resetTradingTarget(rewardAmount);
+                }
+            } else {
+                if (
+                    !_tradingEnabled(token) ||
+                    _amountAvailableForTrading(token) - feeAmount < _minTokenSaleAmounts[token] ||
+                    _amountAvailableForTrading(token) > _minTokenSaleAmountMultiplier * _minTokenSaleAmounts[token] ||
```
barakman: Pretty thin, but casting one of them to `uint256` could be the difference between a successful transaction and one which reverts on overflow just because of this check.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,508 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+                    _totalCollected += feeAmount;
+                } else if (
+                    !_tradingEnabled(token) ||
+                    _amountAvailableForTrading(token) < _minTokenSaleAmounts[token] ||
+                    _auctionPriceIsBelowMinimum(token)
+                ) {
+                    // reset trading for target token
+                    _resetTradingTarget(rewardAmount);
+                }
+            } else {
```
barakman, line 419: You can save the returned value of `_amountAvailableForTrading(token)` into a local variable here, instead of calling that function twice in some cases...

And unfortunately, zero times in other cases, i.e., when `!_tradingEnabled(token)`.

So a possible solution here is something like:
```
else if (!_tradingEnabled(token) {
    // reset trading for token
    _resetTrading(token, rewardAmount);
} else {
    uint128 amount = _amountAvailableForTrading(token);
    if (
        amount - feeAmount < _minTokenSaleAmounts[token] ||
        amount > _minTokenSaleAmountMultiplier * _minTokenSaleAmounts[token] ||
        _auctionPriceIsBelowMinimum(token)
    ) {
        // reset trading for token
        _resetTrading(token, rewardAmount);
    }
}
```

contracts/vortex/CarbonVortex.sol
barakman: Following the code in full (i.e., not just the diffs), I now think that the cast here is undesired to begin with.
Why leave the potential of reverting just because the balance is for some reason larger than 128 bits?
So instead, this function can just return `uint256`.
It is used only for checks, i.e., not for actual assignment of values or anything like that.
So applying this should not cause any issue other than a bit of a headache in casting the other operand in each of these checks to `uint256`.

barakman (changes requested):

barakman (commented):
You should probably use `extend-function-calc-exp-decay-input-range` as the base branch here (and perhaps also pull from it or rebase on top of it), so that the diffs related to that branch don't show up on this PR.

contracts/utility/ExpDecayMath.sol
ivanzhelyazkov: merged with the latest changes

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: the contract is going to be redeployed, so no need to preserve old variables

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: `onlyAdmin` doesn't take care of reentrancy

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: easier to read if it's extracted in a separate function, 
also best to allocate rewards at the end of all state changes (standard security pattern)

test/math/ExpDecayMath.ts
ivanzhelyazkov: merged with the latest changes

test/math/ExpDecayMath.ts
ivanzhelyazkov: merged with the latest changes

test/math/ExpDecayMath.ts
ivanzhelyazkov: merged with the latest changes

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: known issue, also what does this have to do with the current pr ?

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: imo it's better to revert with overflow if the contract has > uint128 token balance.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: gas savings are not worth it vs the readability imo
refunding is better to be done after user has sent the funds, again standard solidity practice

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: we do it 10 lines below

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: gas savings are not worth it vs the readability imo
refunding is better to be done after user has sent the funds, again standard solidity practice

contracts/vortex/CarbonVortex.sol
barakman: When you (theoretically) reenter the function, it has to be from some contract, and not from the admin account (even if that admin account happens to be a contract by itself, though I guess that this can be argued against, so feel free to resolve this comment if you're not convinced).

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: better to revert if any of the numbers are even close to uint128 max value.

contracts/vortex/CarbonVortex.sol
barakman: See [this comment](https://github.com/bancorprotocol/carbon-contracts/pull/140#discussion_r1559754821).

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: _amountAvailableForTrading(token) should always be >= feeAmount

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: no need to continue the auction at this stage, since price is 0, even without reverting the exp decay calculation.

contracts/vortex/interfaces/ICarbonVortex.sol
ivanzhelyazkov: fixed - also added indexed to the tokens array

contracts/vortex/interfaces/ICarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: right, flipped the order of the checks. still left the separate if() check, think it's easier to understand

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: added safeCast
dealing only with uint128 max values for token balances and amounts in the entire code. Doesn't make sense to support balances > uint128

test/forge/VortexTestCaseParser.t.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: extracted the shared logic in `_resetAuction`

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed - extracted the _amountAvailableForTrading(token) in a separate variable. This means only the first auction (only once) for each token will be a bit more expensive, but all later resets are going to have 100 gas saved if the second function call is reached

contracts/vortex/CarbonVortex.sol
barakman: So what exactly are you checking `timeElapsed / currentPriceDecayHalfLife >= 128` for?

I can say that it looks compatible with the restriction that I've **previously** mentioned in the ExpDecay library documentation. But I've updated that restriction since, and I suspect that you might be relying on the previous definition of it.

contracts/vortex/CarbonVortex.sol
barakman: As far as I understand, my suggestion still abides the refunding-rule that you've mentioned.

contracts/vortex/CarbonVortex.sol
barakman: As far as I understand, my suggestion still abides the refunding-rule that you've mentioned.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,502 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
         uint256 len = tokens.length;
 
-        // allocate balances array for the tokens
-        uint256[] memory balances = new uint256[](len);
+        // allocate array for the fee amounts for the tokens
+        uint256[] memory feeAmounts = new uint256[](len);
         // allocate array for the reward amounts for caller
         uint256[] memory rewardAmounts = new uint256[](len);
         // cache rewardsPPM to save gas
         uint256 rewardsPercentage = _rewardsPPM;
```
barakman: That's PPM, not percentage.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,502 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+            newMinTokenSaleAmount: newMinTokenSaleAmount
+        });
     }
 
-    function _setRewardsPPM(uint256 newRewardsPPM) private {
-        uint256 prevRewardsPPM = _rewardsPPM;
+    function _setRewardsPPM(uint32 newRewardsPPM) private {
+        uint32 prevRewardsPPM = _rewardsPPM;
 
         // return if the rewards percentage PPM is the same
```
barakman: Should remove `percentage` in the comment

contracts/vortex/interfaces/ICarbonVortex.sol
```diff
@@ -11,50 +11,186 @@ interface ICarbonVortex is IUpgradeable {
...
+     */
+    event MaxTargetTokenSaleAmountUpdated(uint128 prevTargetTokenSaleAmount, uint128 newTargetTokenSaleAmount);
+
+    /**
+     * @notice triggered when the min token sale amount is updated
+     */
+    event MinTokenSaleAmountUpdated(Token indexed token, uint128 prevMinTokenSaleAmount, uint128 newMinTokenSaleAmount);
 
     /**
      * @dev returns the rewards percentage ppm
```
barakman: Should remove `percentage` in the comment

contracts/vortex/CarbonVortex.sol
```diff
@@ -3,58 +3,102 @@ pragma solidity 0.8.19;
...
-    uint256 private _totalBurned;
+    // first token for swapping
+    Token private immutable _targetToken;
+    // second (optional) token for swapping
+    Token private immutable _finalTargetToken;
+
+    // total target (if no finalTarget token is defined) / finalTarget tokens collected in transferAddress
+    uint256 private _totalCollected;
 
     // rewards percentage
```
barakman: That's PPM, not percentage.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,502 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+                ) {
+                    // reset trading for target token
+                    _resetTradingTarget(rewardAmount);
+                }
+            } else {
+                uint128 tradingAmount = _amountAvailableForTrading(token);
+                if (
+                    !_tradingEnabled(token) ||
+                    tradingAmount - feeAmount < _minTokenSaleAmounts[token] ||
+                    tradingAmount > _minTokenSaleAmountMultiplier * _minTokenSaleAmounts[token] ||
```
barakman, line 424: Since `_minTokenSaleAmountMultiplier` is set to 4 by default, you are essentially restricting the value of `_minTokenSaleAmounts[token]` to 126 bits.

Any value larger than that would cause this transaction to revert on overflow in the expression:
```
_minTokenSaleAmountMultiplier * _minTokenSaleAmounts[token]
```
And this might of course become worse if the admin increases the value of `_minTokenSaleAmountMultiplier`.

My suggestion here is therefore to cast one of the operands in that expression to `uint256`.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,502 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+
+        // check if remaining target token sale amount is below the min target token sale amount
+        if (_targetTokenSaleAmount.current < _minTokenSaleAmounts[_targetToken]) {
+            // top up the target token sale amount
+            _targetTokenSaleAmount.current = Math
+                .min(_targetToken.balanceOf(address(this)), _targetTokenSaleAmount.initial)
+                .toUint128();
+            // reset the price to price * priceResetMultiplier and restart trading
+            Price memory price = tokenPrice(_targetToken);
+            price.sourceAmount = price.sourceAmount * _priceResetMultiplier;
```
barakman: Since `_priceResetMultiplier` is set to 2 by default, you are essentially restricting the value of `price.sourceAmount` to 127 bits.

Any value larger than that would cause this transaction to revert on overflow in the expression:
```
price.sourceAmount * _priceResetMultiplier
```
And this might of course become worse if the admin increases the value of `_priceResetMultiplier`.

My suggestion here is therefore to cast one of the operands in that expression to `uint256`.

contracts/vortex/CarbonVortex.sol
```diff
@@ -81,7 +125,22 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
      * @dev performs contract-specific initialization
      */
     function __CarbonVortex_init_unchained() internal onlyInitializing {
-        _setRewardsPPM(100_000);
+        // set rewards ppm to 5000
+        _setRewardsPPM(5000);
+        // set price reset multiplier to 2x
+        _setPriceResetMultiplier(2);
```
barakman, line 131: Isn't this 2 by definition (of the product)?
If yes, then we can just use a (non-user-configurable) constant.

barakman (changes requested):
I've left a few semantic comments, and a couple of important ones related to the "uint128 stuff", so I'd like to emphasize once again:
I think that we should have as little as possible "uint128 stuff" in the code, meaning that everything stays as uint256 right up to the point when we need to store it somewhere.

contracts/helpers/TestVault.sol
```diff
@@ -0,0 +1,81 @@
...
+        _setRoleAdmin(ROLE_ADMIN, ROLE_ADMIN);
+        _setRoleAdmin(ROLE_ASSET_MANAGER, ROLE_ADMIN);
+        // allow the deployer to initially be the admin of the contract
+        _setupRole(ROLE_ADMIN, msg.sender);
+    }
+
+    // solhint-enable func-name-mixedcase
+
+    /**
+     * @dev authorize the contract to receive the native token
```
lbeder, line 42: It's unrelated to the PR, but in general `@dev` should be only on constructors, modifiers, and private/internal functions, while `@notice` on public/external functions, events, etc.

hardhat.config.ts
```diff
@@ -496,6 +492,14 @@ const config: HardhatUserConfig = {
         ]
     },
 
+    paths: {
+        sources: "./contracts",
```
lbeder: Why are these necessary?

test/math/ExpDecayMath.ts
```diff
@@ -33,7 +33,7 @@ describe('ExpDecayMath', () => {
                 });
             } else {
                 await expect(expDecayMath.calcExpDecay(ethAmount, timeElapsed, halfLife)).to.revertedWithError(
-                    'Overflow'
+                    'panic code 0x11'
```
lbeder, line 36: This can be refactored to `revertedWithPanic` from [@nomicfoundation/hardhat-chai-matchers](https://www.npmjs.com/package/@nomicfoundation/hardhat-chai-matchers) (available through the toolbox or explicitly)

contracts/vortex/interfaces/ICarbonVortex.sol
```diff
@@ -11,50 +11,186 @@ interface ICarbonVortex is IUpgradeable {
...
 
     /**
-     * @dev triggered after a successful burn is executed
+     * @notice triggered after a successful trade is executed
      */
-    event TokensBurned(address indexed caller, Token[] tokens, uint256[] rewardAmounts, uint256 burnAmount);
+    event TokenTraded(address indexed caller, Token indexed token, uint128 sourceAmount, uint128 targetAmount);
 
     /**
      * @dev triggered when the rewards ppm are updated
```
lbeder: E.g., should be `@notice`

contracts/vortex/CarbonVortex.sol
```diff
@@ -3,58 +3,102 @@ pragma solidity 0.8.19;
...
+    // min token sale amounts - resets the token price if below this amount after a call to execute
+    // resets the current sale amount if below this amount after a trade (for target token)
+    mapping(Token token => uint128 _minTokenSaleAmount) private _minTokenSaleAmounts;
+
+    // initial and current target token sale amount - for targetToken->finalTargetToken trades
+    SaleAmount private _targetTokenSaleAmount;
 
     // upgrade forward-compatibility storage gap
-    uint256[MAX_GAP - 2] private __gap;
+    uint256[MAX_GAP - 7] private __gap;
```
lbeder, line 81: I think that it should be 8. Did you verify by printing the storage layout? Also, the forge suit doesn't have an upgradeable storage verification test, like we had in hardhat, right?

contracts/vortex/CarbonVortex.sol
```diff
@@ -81,7 +125,22 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
      * @dev performs contract-specific initialization
      */
     function __CarbonVortex_init_unchained() internal onlyInitializing {
-        _setRewardsPPM(100_000);
+        // set rewards ppm to 5000
```
lbeder: `PPM`?

contracts/vortex/CarbonVortex.sol
```diff
@@ -81,7 +125,22 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+        // set min token sale amount multiplier to 4x
+        _setMinTokenSaleAmountMultiplier(4);
+        // set price decay half-life to 12 hours
+        _setPriceDecayHalfLife(12 hours);
+        // set target token price decay half-life to 12 hours
+        _setTargetTokenPriceDecayHalfLife(12 hours);
+        // set target token price decay half-life to 10 days
+        _setTargetTokenPriceDecayHalfLifeOnReset(10 days);
+        // set initial target token sale amount to 100 eth
+        _setMaxTargetTokenSaleAmount(100 ether);
```
lbeder: This assumes that the target token is ETH, while it's actually configurable. Perhaps it should be something like `100 * 10 ** _targetToke.decimals()`?

contracts/vortex/CarbonVortex.sol
```diff
@@ -81,7 +125,22 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+        // set min target token sale amount to 10 eth
+        _setMinTokenSaleAmount(_targetToken, 10 ether);
```
lbeder: Same here

contracts/vortex/CarbonVortex.sol
```diff
@@ -105,31 +172,171 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+    function withdrawFunds(
+        Token[] calldata tokens,
+        address payable target,
+        uint256[] calldata amounts
+    ) external validAddress(target) validateTokens(tokens) nonReentrant onlyAdmin {
+        uint256 len = tokens.length;
+        if (len != amounts.length) {
+            revert InvalidAmountLength();
+        }
+        for (uint256 i = 0; i < len; i = uncheckedInc(i)) {
```
lbeder, line 303: It's a good opportunity to upgrade to Solidity 0.8.23 (since it's going to be a new deployment) and also avoid `uncheckedInc` in this contract

contracts/vortex/CarbonVortex.sol
```diff
@@ -105,31 +172,171 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+            // safe due to nonReentrant modifier (forwards all available gas in case of ETH)
+            tokens[i].unsafeTransfer(target, amounts[i]);
+        }
+
+        emit FundsWithdrawn({ tokens: tokens, caller: msg.sender, target: target, amounts: amounts });
```
lbeder, line 308: This will be emitted even if `tokens.length` is 0

contracts/vortex/CarbonVortex.sol
```diff
@@ -105,31 +172,171 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+        if (address(_carbonController) != address(0)) {
+            totalFees += _carbonController.accumulatedFees(token);
+        }
+        if (address(_vault) != address(0)) {
+            totalFees += token.balanceOf(address(_vault));
+        }
+        if (address(_oldVortex) != address(0)) {
+            totalFees += token.balanceOf(address(_oldVortex));
+        }
+        return totalFees + token.balanceOf(address(this));
```
lbeder, line 339: Note that `token` can be a rouge token, and this `balanceOf` can be used for a read-only reentrancy (i.e., attacking another contract). No need to address it, IMHO.

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,502 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
-        // withdraw fees, load balances and reward amounts
+        // withdraw fees from carbon, vault and old vortex
         for (uint256 i = 0; i < len; i = uncheckedInc(i)) {
+            Token token = tokens[i];
             // withdraw token fees
-            _carbonController.withdrawFees(tokens[i], type(uint256).max, address(this));
-            // get token balance
-            balances[i] = tokens[i].balanceOf(address(this));
+            uint256 totalFeeAmount = 0;
+            if (address(_carbonController) != address(0)) {
```
lbeder: Perhaps these checked can be cached into a boolean outside of the loop

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,502 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
                 continue;
             }
-
-            // get trade amount for token
-            uint256 tradeAmount;
-            unchecked {
-                // safe because balances[i] >= rewardAmounts[i]
-                tradeAmount = balances[i] - rewardAmounts[i];
+            // skip token if no fees have accumulated or token pair is disabled
+            if (totalFeeAmount == 0 || _disabledPairs[token]) {
```
lbeder, line 397: If a token is disabled, it would still withdraw it above and just do nothing. Perhaps a disabled token should be just ignored (or reverted) altogether?

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,502 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+    function _sellTargetTokenForToken(Token token, uint128 targetAmount) private returns (uint128) {
+        uint128 sourceAmount = expectedTradeInput(token, targetAmount);
+        // revert if trade requires 0 target token
+        if (sourceAmount == 0) {
+            revert InvalidTrade();
+        }
+        // check enough target token (if target token is native) has been sent for the trade
+        if (_targetToken == NATIVE_TOKEN && msg.value < sourceAmount) {
+            revert InsufficientNativeTokenSent();
+        } else {
```
lbeder: The `else` can be refactored away

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,109 +345,502 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+        // revert if trade requires 0 finalTarget tokens
+        if (sourceAmount == 0) {
+            revert InvalidTrade();
+        }
+
+        // check enough final target token (if final target token is native) has been sent for the trade
+        if (_finalTargetToken == NATIVE_TOKEN) {
+            if (msg.value < sourceAmount) {
+                revert InsufficientNativeTokenSent();
+            } else {
```
lbeder: Same here

contracts/vortex/CarbonVortex.sol
```diff
@@ -105,31 +172,171 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+    }
+
+    /**
+     * @notice sets if trading is enabled or disabled for a token
+     *
+     * requirements:
+     *
+     * - the caller must be the admin of the contract
+     */
+    function disablePair(Token token) external onlyAdmin {
```
lbeder: This should be explicit (instead of triggering). Triggering is a bad UX due to human error (e.g., the DAO sends multiple transactions accidentally and not realizing that the state won't end up as intended)

contracts/vortex/CarbonVortex.sol
```diff
@@ -252,40 +852,125 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+        // get time elapsed since trading was enabled
+        uint32 timeElapsed = uint32(block.timestamp) - tradingStartTime;
+        // get the halflife for the token
+        uint32 currentPriceDecayHalfLife = token == _targetToken ? _targetTokenPriceDecayHalfLife : _priceDecayHalfLife;
+        // check if the maximum amount of halflifes have been reached
+        if (timeElapsed / currentPriceDecayHalfLife >= 128) {
+            return true;
+        } else {
+            return false;
         }
```
lbeder: Nitpick: you can return the condition directly

lbeder (changes requested):
This is a preliminary review, but FYI

contracts/vortex/CarbonVortex.sol
barakman: Something to keep in mind here is that some early-phase (2017ish) token contracts have implemented `DECIMALS` or `Decimals` instead of `decimals`, as far as I remember.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: it should be configurable

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: I'd argue the point that _minTokenSaleAmounts are essentially token balances - and for almost all tokens even uint112 is enough to hold their balances (for example, Uniswap uses uint112 to represent token reserves)

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: reallistically speaking, the price.sourceAmount for target tokens will be at market price at the point of reset - no one will be trading at a price of 2^127.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/interfaces/ICarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Verified by printing the layout - the last storage var - `_targetTokenSaleAmount` takes up only one slot (2 uint128).
The forge test suite doesn't have the storage verification tests, but test:deploy has it (tenderly deployment test)

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: points per million

lbeder: I meant that it should be "PPM" and not "ppm"

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: these values can be configured by the admin on deployment on each chain

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: these values can be configured by the admin on deployment on each chain

contracts/vortex/CarbonVortex.sol
lbeder: My point is that the defaults can be buggy

contracts/vortex/CarbonVortex.sol
lbeder: My point is that the defaults can be buggy

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Tokens get disabled for trading only - it should still be withdrawable (product spec)

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: `_validateTokens` has a check `if (tokens.length == 0)` which will revert the call.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: yes, aware of the issue - discussed internally with Yudi if we shouldn't refactor the `withdrawFunds` method to return the withdrawn amount in a similar way to the controller - but decided not to at this stage

test/math/ExpDecayMath.ts
ivanzhelyazkov: should be addressed by Barak in his PR related to the exp decay math - and merged later into this branch

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: agree, fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/interfaces/ICarbonVortex.sol
ivanzhelyazkov: fixed

hardhat.config.ts
ivanzhelyazkov: removed the default paths - only left the deploy dir

contracts/helpers/TestVault.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: there are a lot of imports with version 0.8.19 (tests as well) - they'll have to be changed as well - I think we should make it a general refactor, if at all.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: changed to `100 * 10 ** _targetToken.decimals()`

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: changed to `10 * 10 ** _targetToken.decimals()`

barakman (approved):

test/math/ExpDecayMath.ts
lbeder: @barakman FYI

barakman: See https://github.com/bancorprotocol/carbon-contracts/pull/139#discussion_r1554954444 (decided to leave it as is and resolved the comment).

contracts/vortex/CarbonVortex.sol
```diff
@@ -3,58 +3,102 @@ pragma solidity 0.8.19;
...
 /**
- * @dev CarbonVortex contract
+ * @notice CarbonVortex contract
  */
 contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable, Utils {
     using Address for address payable;
+    using SafeCast for uint256;
+
+    uint128 private constant INITIAL_PRICE_SOURCE_AMOUNT = type(uint128).max;
+    uint128 private constant INITIAL_PRICE_TARGET_AMOUNT = 1e12;
```
lbeder, line 28: What's the reasoning behind this value?

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,111 +345,505 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+
+        // check if remaining target token sale amount is below the min target token sale amount
+        if (_targetTokenSaleAmount.current < _minTokenSaleAmounts[_targetToken]) {
+            // top up the target token sale amount
+            _targetTokenSaleAmount.current = Math
+                .min(_targetToken.balanceOf(address(this)), _targetTokenSaleAmount.initial)
+                .toUint128();
+            // reset the price to price * priceResetMultiplier and restart trading
+            Price memory price = tokenPrice(_targetToken);
+            price.sourceAmount = price.sourceAmount * _priceResetMultiplier;
```
lbeder: Can be *=

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,111 +345,505 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+
+        // check enough final target token (if final target token is native) has been sent for the trade
+        if (_finalTargetToken == NATIVE_TOKEN) {
+            if (msg.value < sourceAmount) {
+                revert InsufficientNativeTokenSent();
+            }
+            payable(_transferAddress).sendValue(sourceAmount);
+        } else {
+            // transfer the tokens from the user to the _transferAddress
+            _finalTargetToken.safeTransferFrom(msg.sender, _transferAddress, sourceAmount);
```
lbeder, line 664: `_finalTargetToken` can be `0x0` and then `_sellFinalTargetForTarget` will revert, in which case `trade` will never work

test/math/ExpDecayMath.ts
lbeder: @ivanzhelyazkov so you can go ahead and refactor this yourself?

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: if `finalTargetToken` is `0x0`, we don't want `sellFinalTargetForTarget` to work - we directly transfer the `targetToken` to the `transferAddress` when it's:
1) withdrawn using `execute`
2) traded via `_sellTargetTokenForToken`

test/math/ExpDecayMath.ts
ivanzhelyazkov: I do think here this should be related to a bigger refactor - waffle and hardhat-chai-matchers don't work with each other, and some packages need to be removed - as well as some changes made to the tests themselves - given that there's only a few stress tests which aren't run often, imo this change can be planned as a later refactor to the repo.

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Goal here is to start from the maximum possible auction price to encompass all of the different token prices. Auction starts at max price and progresses slowly via exp decay based on the specified hafllife (currently set to 12 hours).
After a while for each token, market price should be reached and the tokens should be open for arbing if market conditions allow it (profit from the arb exceed the gas costs)

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

contracts/vortex/CarbonVortex.sol
lbeder: So it's ok that `trade` won't work in this case either?

lbeder (approved):

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Yes

contracts/vortex/CarbonVortex.sol
```diff
@@ -138,111 +345,505 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
...
+        uint128 sourceAmount;
+        if (token == _targetToken) {
+            sourceAmount = _sellFinalTargetForTarget(targetAmount);
+        } else {
+            sourceAmount = _sellTargetTokenForToken(token, targetAmount);
+        }
+        emit TokenTraded({ caller: msg.sender, token: token, sourceAmount: sourceAmount, targetAmount: targetAmount });
+    }
+
+    function _sellTargetTokenForToken(Token token, uint128 targetAmount) private returns (uint128) {
```
yudilevi: the contract doesn't sell the target token for the token, it does the opposite. same for the final target function

yudilevi (changes requested):

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed

yudilevi (approved):

PR #141, opened 2024-04-22
Carbon vortex 2.0 deployment scripts

ivanzhelyazkov:
* Add carbon vortex 2.0 deployment scripts
* Update eslint


yudilevi (changes requested):
I think the ProxyAdmin should also be renamed

yudilevi (approved):

lbeder (approved):
Maybe all the `_Old` artifacts can be just removed? I don't see anyone using them like this (instead of taking a previous version) nor can they be referenced from deployment scripts

PR #149, opened 2024-06-14
carbon vortex 2.0 target token min sale amount handling

ivanzhelyazkov:
* Add handling for a below the min sale amount case for the target token during TKN->target token trades
* Add tests

### DIFF
--- a/contracts/vortex/CarbonVortex.sol
+++ b/contracts/vortex/CarbonVortex.sol
@@ -652,6 +652,15 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
             _resetTrading(token, 0);
         }
 
+        // if available target token trading amount is below the min target token sale amount, reset the target token auction
+        if (
+            Token.unwrap(_finalTargetToken) != address(0) &&
+            _amountAvailableForTrading(_targetToken) <
+            _minTokenSaleAmounts[_targetToken] / _minTokenSaleAmountMultiplier
+        ) {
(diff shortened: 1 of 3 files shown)


yudilevi (approved):

contracts/vortex/CarbonVortex.sol
```diff
@@ -652,6 +652,15 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
             _resetTrading(token, 0);
         }
 
+        // if available target token trading amount is below the min target token sale amount, reset the target token auction
+        if (
+            Token.unwrap(_finalTargetToken) != address(0) &&
+            _amountAvailableForTrading(_targetToken) <
+            _minTokenSaleAmounts[_targetToken] / _minTokenSaleAmountMultiplier
```
barakman, line 659: If accuracy is critical here for `_resetTradingTarget` to take place, then you should change the division on the RHS to a multiplication on the LHS (and of course, cast the value returned from `_amountAvailableForTrading` to `uint256`, in order to avoid reverting on overflow).

lbeder (approved):

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: Was thinking about this - but considering the expected values for minTokenSaleAmounts and minTokenSaleMultiplier, I think that RHS division is sufficient

barakman (approved):

PR #156, opened 2024-08-06
Gradient strategies test improvement

barakman:
Extend the gradient-strategies testing to include:
1. Asserting the accuracy of the strategy configuration (requested parameters vs encoded parameters)
2. Asserting the accuracy of function `calcCurrentRate` with respect to the encoded parameters instead of the requested parameters


ivanzhelyazkov (approved):
Looks good !

PR #162, opened 2024-08-21
fix upgradeable contract and health tests

ivanzhelyazkov:
* Fix upgradeable contract versioning issue
* Fix health and deploy tests
* Add postUpgrade test for new deployments

### DIFF
--- a/contracts/utility/Upgradeable.sol
+++ b/contracts/utility/Upgradeable.sol
@@ -83,10 +83,11 @@ abstract contract Upgradeable is IUpgradeable, AccessControlEnumerableUpgradeabl
      */
     function postUpgrade(bytes calldata data) external {
         uint16 initializations = _initializations + 1;
-        if (initializations < version()) {
-            initializations = version();
+        uint16 _version = version();
+        if (initializations < _version) {
+            initializations = _version;
         }
-        if (initializations != version()) {
+        if (initializations != _version) {
             revert AlreadyInitialized();
         }
 
--- a/contracts/utility/Upgradeable.sol
+++ b/contracts/utility/Upgradeable.sol
@@ -81,17 +81,16 @@ abstract contract Upgradeable is IUpgradeable, AccessControlEnumerableUpgradeabl
      *
      * - this must and can be called only once per-upgrade
      */
-    function postUpgrade(bytes calldata data) external {
+    function postUpgrade(bool checkVersion, bytes calldata data) external {
         uint16 initializations = _initializations + 1;
         uint16 _version = version();
-        if (initializations < _version) {
-            initializations = _version;
-        }
-        if (initializations != _version) {
+        if (checkVersion && initializations != _version) {
             revert AlreadyInitialized();
+        } else if (!checkVersion) {
+            initializations = _version;
+        } else {
+            initializations = initializations;
         }
-
-        _initializations = initializations;
 
         _postUpgrade(data);
     }
--- a/contracts/utility/Upgradeable.sol
+++ b/contracts/utility/Upgradeable.sol
@@ -42,7 +42,7 @@ abstract contract Upgradeable is IUpgradeable, AccessControlEnumerableUpgradeabl
      * @dev performs contract-specific initialization
      */
     function __Upgradeable_init_unchained() internal onlyInitializing {
-        _initializations = version() - 1;
+        _initializations = version();
 
         // set up administrative roles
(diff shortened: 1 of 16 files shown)


contracts/utility/Upgradeable.sol
```diff
@@ -83,6 +83,9 @@ abstract contract Upgradeable is IUpgradeable, AccessControlEnumerableUpgradeabl
      */
     function postUpgrade(bytes calldata data) external {
         uint16 initializations = _initializations + 1;
+        if (initializations < version()) {
```
lbeder: You can cache `version()`, but it's not the end of the world

lbeder: The idea was preventing accidental non-incremental version upgrades. The case that you have version 5 and trying to upgrade it to version 8 shouldn't happen, no?

yudilevi: What we're trying to prevent here is someone calling postUpgrade multiple times.

We have multiple different scenarios:
1. CarbonController is deployed with version 2 without the proxy delegate check, and then immediately upgraded with the proxy delegate check (postUpgrade is called)
2. Other contracts are deployed with version 1 and no upgrade (postUpgrade isn't called)
3. Multiple different versions are deployed to different chains, so we don't have to maintain a version number per chain, this can result in a deployment of version 3 to chain X

ivanzhelyazkov: fixed

contracts/utility/Upgradeable.sol
```diff
@@ -81,14 +81,17 @@ abstract contract Upgradeable is IUpgradeable, AccessControlEnumerableUpgradeabl
...
+    function postUpgrade(bool checkVersion, bytes calldata data) external {
         uint16 initializations = _initializations + 1;
-        if (initializations != version()) {
+        uint16 _version = version();
+        if (checkVersion && initializations != _version) {
             revert AlreadyInitialized();
+        } else if (!checkVersion) {
+            initializations = _version;
+        } else {
+            initializations = initializations;
```
lbeder: Should be `_initializations`, no? Also implying of a missing test

contracts/utility/Upgradeable.sol
```diff
@@ -81,14 +81,17 @@ abstract contract Upgradeable is IUpgradeable, AccessControlEnumerableUpgradeabl
...
      * - this must and can be called only once per-upgrade
      */
-    function postUpgrade(bytes calldata data) external {
+    function postUpgrade(bool checkVersion, bytes calldata data) external {
         uint16 initializations = _initializations + 1;
-        if (initializations != version()) {
+        uint16 _version = version();
+        if (checkVersion && initializations != _version) {
             revert AlreadyInitialized();
+        } else if (!checkVersion) {
```
lbeder, line 89: Nitpick: you don't need the `else`, since `revert` is treated as `return`

contracts/utility/Upgradeable.sol
ivanzhelyazkov: fixed

contracts/utility/Upgradeable.sol
ivanzhelyazkov: fixed

contracts/utility/Upgradeable.sol
```diff
@@ -88,10 +88,10 @@ abstract contract Upgradeable is IUpgradeable, AccessControlEnumerableUpgradeabl
             revert AlreadyInitialized();
         } else if (!checkVersion) {
             initializations = _version;
-        } else {
-            initializations = initializations;
         }
 
+        _initializations = initializations;
```
lbeder, line 93: Nitpick: you write twice to `_initializations` (but it's also not the end of the world, since it's called only during an upgrade)

ivanzhelyazkov: only one write I think (we set `initializations` (local var) on line 85, then we write to it on line 90 (the local var), and then we set `_initializations` (the storage var) to `initializations` at the end)

lbeder (approved):

contracts/utility/Upgradeable.sol
yudilevi: very minor, but I still see if/else, though in this specific case I'd do the `!checkVersion` first, so the var isn't checked again in the opposite clause

yudilevi (approved):

PR #168, opened 2024-10-07
carbon vortex - minor fixes

ivanzhelyazkov:
* Fix execute logic for final target token - transfer the funds to the `transferAddress` directly instead of skipping it
* Make `transferAddress` settable
* Add additional logic for `transferAddress` - if set to 0x0, leave funds from execute and trade in the vortex

### DIFF
--- a/contracts/vortex/CarbonVortex.sol
+++ b/contracts/vortex/CarbonVortex.sol
@@ -1033,16 +1033,14 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
     }
 
     function _transferProceeds(Token token, uint256 amount) private {
+        // increment totalCollected amount
+        _totalCollected += amount;
         // if transfer address is 0, proceeds stay in the vortex
         if (_transferAddress == address(0)) {
-            // increment totalCollected amount
-            _totalCollected += amount;
             return;
         }
         // safe due to nonReentrant modifier (forwards all available gas in case of ETH)
         token.unsafeTransfer(_transferAddress, amount);
-        // increment totalCollected amount
-        _totalCollected += amount;
     }
 
     function uncheckedInc(uint256 i) private pure returns (uint256 j) {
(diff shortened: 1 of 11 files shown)


contracts/vortex/CarbonVortex.sol
```diff
@@ -1012,6 +1032,19 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
         return price;
     }
 
+    function _transferProceeds(Token token, uint256 amount) private {
+        // if transfer address is 0, proceeds stay in the vortex
+        if (_transferAddress == address(0)) {
+            // increment totalCollected amount
+            _totalCollected += amount;
```
lbeder: Nitpick: you can call this first and then you can call it once (CEI isn't essential here, but it's your call)

contracts/vortex/CarbonVortex.sol
```diff
@@ -1012,6 +1032,19 @@ contract CarbonVortex is ICarbonVortex, Upgradeable, ReentrancyGuardUpgradeable,
         return price;
     }
 
+    function _transferProceeds(Token token, uint256 amount) private {
+        // if transfer address is 0, proceeds stay in the vortex
+        if (_transferAddress == address(0)) {
```
lbeder, line 1039: Why allowing the funds to stay in the Vortex is a good idea?

yudilevi: It’s actually a requirement. The next phase would be to allow sending them to a bridging contract and then to Ethereum mainnet

contracts/vortex/CarbonVortex.sol
ivanzhelyazkov: fixed, was thinking about it as well

contracts/vortex/CarbonVortex.sol
lbeder: Roger that

lbeder (approved):

yudilevi (approved):

lbeder (approved):

PR #175, opened 2024-12-18
Carbon batch router

ivanzhelyazkov:
* Add CarbonBatcher contract - creates multiple strategies with one transaction
* Add tests
* Add deployment scripts


contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+            Token token = uniqueTokens[i];
+            uint256 amount = amounts[i];
+            if (token.isNative()) {
+                if (txValueLeft < amount) {
+                    revert InsufficientNativeTokenSent();
+                }
+                txValueLeft -= amount;
+                continue;
+            }
+            token.safeTransferFrom(msg.sender, address(this), amount);
```
barakman, line 95: This code is never reached

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+    Order[2] orders;
+}
+
+/**
+ * @dev Contract to batch create carbon controller strategies
+ */
+contract CarbonBatcher is Upgradeable, Utils, ReentrancyGuard, IERC721Receiver {
+    using Address for address payable;
+
+    error InsufficientNativeTokenSent();
```
lbeder: Can this error move to common/shared utils?

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+
+    ICarbonController private immutable carbonController;
```
lbeder: `_carbonController`?

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+    IVoucher private immutable voucher;
```
lbeder: `_voucher`?

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+
+    /**
+     * @dev triggered when tokens have been withdrawn from the carbon batcher
+     */
+    event FundsWithdrawn(Token indexed token, address indexed caller, address indexed target, uint256 amount);
+
+    constructor(
+        ICarbonController _carbonController,
```
lbeder: We don't prefix function arguments with an `_`, but if you did this in order to resolve the conflict with state variables, you can call rename these arguments to `initCarbonController`, and `initVoucher`, for example

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+        IVoucher _voucher
+    ) validAddress(address(_carbonController)) validAddress(address(_voucher)) {
+        carbonController = _carbonController;
+        voucher = _voucher;
+        _disableInitializers();
```
lbeder, line 39: Missing new line?

contracts/token/Token.sol
```diff
@@ -135,6 +137,18 @@ function safeApprove(Token token, address spender, uint256 amount) {
...
+/**
+ * @dev force approves a specific amount of the native token/ERC20 token from a specific holder
+ *
+ * note that the function does not perform any action if the native token is provided
+ */
+function forceApprove(Token token, address spender, uint256 amount) {
+    if (isNative(token)) {
+        return;
+    }
+    toIERC20(token).forceApprove(spender, amount);
```
lbeder, line 150: Missing new line?

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    /**
+     * @dev fully initializes the contract and its parents
+     */
+    function initialize() public initializer {
```
lbeder: Can it be `external`?

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+
+    /**
+     * @inheritdoc Upgradeable
+     */
+    function version() public pure virtual override(Upgradeable) returns (uint16) {
+        return 1;
+    }
+
+    /**
+     * @notice creates several new strategies, returns the strategies id's
```
lbeder: `ids`?

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+    function batchCreate(
+        StrategyData[] calldata strategies
+    ) external payable greaterThanZero(strategies.length) nonReentrant returns (uint256[] memory) {
+        uint256[] memory strategyIds = new uint256[](strategies.length);
+        uint256 txValueLeft = msg.value;
+
+        // extract unique tokens and amounts
+        (Token[] memory uniqueTokens, uint256[] memory amounts) = _extractUniqueTokensAndAmounts(strategies);
+        // transfer funds from user for strategies
+        for (uint256 i = 0; i < uniqueTokens.length; i = uncheckedInc(i)) {
```
lbeder, line 81: Perhaps it's time to upgrade to Solidity >0.8.22? `uncheckedInc(i)` will be no longer needed, for example

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+            Token token = uniqueTokens[i];
+            uint256 amount = amounts[i];
+            if (token.isNative()) {
+                if (txValueLeft < amount) {
+                    revert InsufficientNativeTokenSent();
```
lbeder, line 86: This check isn't strictly needed due to the checked subtraction below

contracts/utility/CarbonBatcher.sol
lbeder, line 91: These can be in an `else` clause to prevent nop call for ETH

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        // create strategies and transfer nfts to user
+        for (uint256 i = 0; i < strategies.length; i = uncheckedInc(i)) {
+            // get tokens for this strategy
+            Token[2] memory tokens = strategies[i].tokens;
+            Order[2] memory orders = strategies[i].orders;
+            // if any of the tokens is native, send this value with the create strategy tx
+            uint256 valueToSend = 0;
+            if (tokens[0].isNative()) {
```
lbeder: You can cache `tokens[0]`and `tokens[1]`

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+                valueToSend = orders[0].y;
+            } else if (tokens[1].isNative()) {
+                valueToSend = orders[1].y;
+            }
+
+            // create strategy on carbon
+            strategyIds[i] = carbonController.createStrategy{ value: valueToSend }(tokens[0], tokens[1], orders);
+            // transfer nft to user
+            voucher.safeTransferFrom(address(this), msg.sender, strategyIds[i], "");
```
lbeder: Hmmm.. maybe `createStrategy` can receive an `receipient` or `owner` argument and mint the voucher to it directly? This will prevent the additional transfer.

lbeder: You can cache `strategyIds[i]` and assign it only later

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+     *
+     * requirements:
+     *
+     * - the caller is admin of the contract
+     */
+    function withdrawFunds(
+        Token token,
+        address payable target,
+        uint256 amount
+    ) external validAddress(target) onlyAdmin nonReentrant {
```
lbeder, line 137: Why is this function required?

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+            return;
+        }
+
+        // forwards all available gas in case of ETH
+        token.unsafeTransfer(target, amount);
+
+        emit FundsWithdrawn({ token: token, caller: msg.sender, target: target, amount: amount });
+    }
+
+    function onERC721Received(address, address, uint256, bytes calldata) external pure returns (bytes4) {
```
lbeder, line 151: Missing NATSPEC?

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+
+        for (uint256 i = 0; i < uniqueCount; i = uncheckedInc(i)) {
+            uniqueTokens[i] = tempUniqueTokens[i];
+            amounts[i] = tempAmounts[i];
+        }
+
+        return (uniqueTokens, amounts);
+    }
+
+    function _findInArray(Token element, Token[] memory array, uint256 arrayLength) private pure returns (uint256) {
```
lbeder, line 203: Missing NATSPEC

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,224 @@
...
+            return;
+        }
+        uint256 allowance = token.toIERC20().allowance(address(this), address(carbonController));
+        if (allowance < inputAmount) {
+            // increase allowance to the max amount if allowance < inputAmount
+            token.forceApprove(address(carbonController), type(uint256).max);
+        }
+    }
+
+    function uncheckedInc(uint256 i) private pure returns (uint256 j) {
```
lbeder, line 229: Missing NATSPEC

lbeder, line 229: Should it moved to a common/shared utilities?

lbeder (changes requested):

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: need to either change all contracts to >0.8.22, which would probably require some other changes and would make the currently deployed carbon controller, vortex versions not work properly or would require adding an additional compiler for this repo, probably with additional OZ versions, etc.
can be considered probably apart from this PR as a separate general repo upgrade

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: prefer to have a specific error for this case to ease debugging

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: in ETH's case, we continue to the next token in the array on line 93 (or revert if unsufficient eth has been sent) - so we shouldn't hit lines 95 and 96.

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: Well, that would require changing the CarbonController's code, but it's already immutable - so I don't think it's possible at this stage

ivanzhelyazkov: What do you mean here ?

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: According to the spec, it's a withdraw function to extract tokens - safety function

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: Added

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: Added

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: Added

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: fixed

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: fixed

contracts/token/Token.sol
ivanzhelyazkov: fixed

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: fixed

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: fixed

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: fixed

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: fixed

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: fixed, moved to Utils

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: maybe as a part of a later general upgrade to the entire repo ? (if we upgrade the solidity version we won't even need it)

contracts/utility/CarbonBatcher.sol
lbeder: AFAIK, versions 0.8.23 to 0.8.28 are extremely portable, but this can indeed wait. I'd recommend upgrading eventually, though

contracts/utility/CarbonBatcher.sol
lbeder: Perhaps a version check? It's a bummer to exclude this gas optimization forever at this point. Pinging @yudilevi

contracts/utility/CarbonBatcher.sol
lbeder: Gotcha

contracts/utility/CarbonBatcher.sol
lbeder: Alrighty

contracts/utility/CarbonBatcher.sol
lbeder: I've meant that you can store the strategy id in a local variable and only add it to the array later

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: done

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: done

lbeder (approved):

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,236 @@
...
+    ICarbonController private immutable _carbonController;
+    IVoucher private immutable _voucher;
+
+    /**
+     * @dev triggered when tokens have been withdrawn from the carbon batcher
+     */
+    event FundsWithdrawn(Token indexed token, address indexed caller, address indexed target, uint256 amount);
+
+    constructor(
+        ICarbonController carbonControllerInit,
```
yudilevi: can get the controller from the voucher

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,236 @@
...
+        uint256 txValueLeft = msg.value;
+
+        // extract unique tokens and amounts
+        (Token[] memory uniqueTokens, uint256[] memory amounts) = _extractUniqueTokensAndAmounts(strategies);
+        // transfer funds from user for strategies
+        for (uint256 i = 0; i < uniqueTokens.length; i = uncheckedInc(i)) {
+            Token token = uniqueTokens[i];
+            uint256 amount = amounts[i];
+            if (token.isNative()) {
+                if (txValueLeft < amount) {
```
yudilevi, line 85: why check here instead of transferring the entire amount and letting carbon deal with it?

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,236 @@
...
+
+            // create strategy on carbon
+            uint256 strategyId = _carbonController.createStrategy{ value: valueToSend }(token0, token1, orders);
+            // transfer nft to user
+            _voucher.safeTransferFrom(address(this), msg.sender, strategyId, "");
+            // assign strategy id
+            strategyIds[i] = strategyId;
+        }
+        // refund user any remaining native token
+        if (txValueLeft > 0) {
```
yudilevi, line 117: can refund any remaining balance as a safety mechanism, as carbon might decide to refund some amount as well, and in that case this contract will have some residue

contracts/utility/CarbonBatcher.sol
```diff
@@ -0,0 +1,236 @@
...
+            // forwards all available gas
+            payable(msg.sender).sendValue(txValueLeft);
+        }
+
+        return strategyIds;
+    }
+
+    /**
+     * @notice withdraws funds held by the contract and sends them to an account
```
yudilevi, line 126: maybe also note that this is a safety mechanism but that in general it should never be necessary.

yudilevi (changes requested):

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: save some gas if user forces the tx

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: fixed

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: added

contracts/utility/CarbonBatcher.sol
ivanzhelyazkov: if we send `y` value with the strategy creation, carbon shouldn't refund extra funds - so should be only the leftover sent by the user to the batcher

yudilevi (approved):

PR #188, opened 2025-08-07
deployment scripts - add support deployment type

ivanzhelyazkov:
* Add a `type` flag to the deployment scripts
* If type == network, deploy the full configuration (ProxyAdmin, Carbon Controller, Voucher, Vortex and Batcher)
* If type == support, deploy only the ProxyAdmin and Vortex contracts


deployments/run-network.sh
```diff
@@ -1,10 +1,38 @@
 #!/bin/bash
```
lbeder, line 1: I'd recommend adding `set -e` to all the scripts

lbeder (approved):

yudilevi (approved):

deployments/run-network.sh
ivanzhelyazkov: added

